PR #4, opened 2022-06-15
Add Digital Signature Parameters To Function Stub

LuisOsta:


fjarri:
It's hard for me to judge this, since I don't know how tightly `ssi` is supposed to be coupled with `monetae-sdk`, and what exactly will be happening in those methods. But as is it looks good to me.

fjarri (approved):

sKudryashov (approved):

Ryanmtate (approved):

Ryanmtate:
> It's hard for me to judge this, since I don't know how tightly `ssi` is supposed to be coupled with `monetae-sdk`, and what exactly will be happening in those methods. But as is it looks good to me.

Ideally, we keep the interaction with monetae-sdk at an abstracted level. The `signer`, `Signature` and `Verifier` traits help us to achieve this level of abstraction.

SSI does not need to know anything about monetae for it to work.

PR #5, opened 2022-06-17
Verifiable credentials creation

sKudryashov:


src/credential.rs
```diff
@@ -0,0 +1,61 @@
...
+
+pub struct CredentialProof {
+	proof_type: String,
+	created: String,
+	verification_method: String,
+	proof_purpose: String,
+	proof_value: String,
+}
+
+pub struct VerifiablePresentation {
```
LuisOsta, line 57: This is for service level correctness so we shouldn't need it for the work that we need to do here. For creating VPs the only thing that we would need from the user is basically an array of the info we need to create a VP

src/credential.rs
```diff
@@ -0,0 +1,61 @@
...
+	id: String,
+    cred_type: String,
+	issuer: String,
+	issuance_date: String,
+	subject: CredentialSubject,
+	proof: *mut CredentialProof,
+}
+
+impl <'a> VerifiableCredential {
+    pub fn init(cred_type: &'a str, cred_subject: serde_json::Value) -> Self  {
```
LuisOsta, line 21: The `init` of `VerifiablePresentation` will have an array of the same parameters. So an array of cred types and cred subjects

LuisOsta: You can use this as a reference point of what the VP creation will need - https://www.w3.org/TR/vc-data-model/#presentations-0. There are some things that we can self determine (such as issuance date) but everything that we can't get ourselves will have to be passed in

core/src/credential.rs
```diff
@@ -0,0 +1,97 @@
...
+pub(crate) const CRED_TYPE_BANK_CARD: str = "BankCard";
+
+type CredentialSubject = serde_json::Value;
+
+enum CredType {
+    PermanentResidentCard(String),
+    BankCard(String)
+}
+
+pub trait VCCredential {
```
LuisOsta, line 20: I think we should call this `DocumentBuilder`

core/src/credential.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn getCredType (&self) -> String;
+    fn getSubject (&self) -> String;
+    fn getContext (&self) -> [String];
+}
+
+impl VCCredential for VerifiableCredential<'_> {
```
LuisOsta, line 26: This is unnecessary, what we should do is have the trait have the default implementations

LuisOsta: What I meant is that we should move `create_credential` inside the `DocumentBuilder`. It would live alongside `create_presentation`.

Then `create_credential` can pass everything needed to create a credential to whatever function we define in `credential.rs`. Does that make sense?

core/src/credential.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct VerifiableCredential <'a> {
+    #[serde(rename = "@context")]
+	context:  VCContext,
+	id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    cred_type:  Option<CredType>,
```
LuisOsta, line 44: I don't think we should have an enum here, since users can override the cred type we don't have those kinds of guarantees. Also why is this an `Option<T>`?

core/src/credential.rs
```diff
@@ -0,0 +1,97 @@
...
+	issuer: String,
+	issuance_date: String,
+	subject: CredentialSubject,
+	proof: *mut CredentialProof,
+}
+
+impl <'a> VerifiableCredential <'a>  {
+    pub fn init(cred_type: String, cred_subject: serde_json::Value, issuer: &'a str, id: &'a str) -> Self  {
```
LuisOsta, line 52: I think `new` might be more standard in Rust than `init` but I'm unsure

core/src/credential.rs
```diff
@@ -0,0 +1,97 @@
+
+#![allow(unused_variables)]
+#![allow(dead_code)]
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+type VCContext = [String];
+
+pub(crate) const JSON_LD_CONTEXT:  VCContext = ["https://www.w3.org/2018/credentials/v1", "https://www.w3.org/2018/credentials/examples/v1"];
```
Ryanmtate, line 9: May be worth converting these to lazy statics to load them when you need them, e.g.:

```rust
lazy_static! {
    pub const CREDENTIAL_V1_JSON_LD_CONTEXT: &str = "https://www.w3.org/2018/credentials/v1"
}
```

fkim7: so for PermanentResidentCard credType would that entail adding a citizenship const "https://w3id.org/citizenship/v1"?

rtsuk: I don't believe there is a load time cost for a const &str or an array of them. I'm pretty sure it is baked into the binary image.

Ryanmtate: Maybe another question whether the array is the best structure for these-- but not sure.

core/src/credential.rs
```diff
@@ -0,0 +1,97 @@
+
+#![allow(unused_variables)]
+#![allow(dead_code)]
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+type VCContext = [String];
```
Ryanmtate, line 7: should this be `[&'static str]` instead?

LuisOsta: This is going to be changed as whole so I don't think we'll be keeping it

core/src/credential.rs
rtsuk: Indeed it is all at build time https://rust.godbolt.org/z/Goajr5a4W

core/src/credential.rs
sKudryashov: The Option<T> here was to handle the case where there is no cred type, but from our morning conversation i realize that it won't be the case anyway, we are returning error if cred_type is redefined and empty.

core/src/credential.rs
sKudryashov: I agree that `new` is more standard in terms of creating new struct, but motivation here is to underline that there is some object initialization logic, it is more like a factory then a constructor. Anyways, i'm happy to change according to language conventions. Maybe @Ryanmtate team has some PR review checklist that could help us in terms of language conventions and self-review?

core/src/credential.rs
sKudryashov: Yes, [&'static str] is better here.

core/src/credential.rs
```diff
@@ -0,0 +1,97 @@
...
+pub struct VerifiableCredential <'a> {
+    #[serde(rename = "@context")]
+	context:  VCContext,
+	id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    cred_type:  Option<CredType>,
+	issuer: String,
+	issuance_date: String,
+	subject: CredentialSubject,
+	proof: *mut CredentialProof,
```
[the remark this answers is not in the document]
sKudryashov: Maybe make it an optional Option<CredentialProof> with `#[serde(skip_serializing_if = "Option::is_none")]`  pragma?

core/src/credential.rs
```diff
@@ -0,0 +1,145 @@
...
+                      "@protected": true,
+                      "id": "@id",
+                      "type": "@type",
+                      "description": "http://schema.org/description",
+                      "name": "http://schema.org/name",
+                      "identifier": "http://schema.org/identifier",
+                      "image": {"@id": "http://schema.org/image", "@type": "@id"}
+                    }
+                  }"###),
+                (CRED_TYPE_BANK_CARD, r###"{
```
fkim7, line 81: These attributes look like they belong to permanent resident card https://w3c-ccg.github.io/citizenship-vocab/#:~:text=3.-,The%20Citizenship%20Vocabulary,-This%20vocabulary%20assumes

Note the bankcard VC we have defined so far as a placeholder (Subject to change, but the w3c citizenship/permanent resident card above is not) is defined here https://www.notion.so/Verifiable-Credentials-Adapter-MVP-Design-56e1ea273c054fb39cc27ea1c8fcd744#8628511fbb99474aaa3617ee9eb00846

Cargo.toml
```diff
@@ -3,3 +3,15 @@ members = [
...
 	"registry-resolver"
 ]
+[package]
+name = "ssi"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
LuisOsta, line 13: Why are we adding dependencies here? This is the top level `Cargo.toml` we should add the dependencies for the core functionality in the `core/Cargo.toml` file

Cargo.toml
```diff
@@ -3,3 +3,15 @@ members = [
 	"core",
 	"registry-resolver"
 ]
+[package]
```
LuisOsta, line 6: This should be removed as the `ssi` package will be the one under the `core` folder. Unless there's something I'm missing about how Cargo works (could be I'm not as familiar)

core/src/lib.rs
```diff
@@ -14,23 +23,35 @@ pub trait DIDResolver {
...
-pub fn create_presentation(
-    _creds: Vec<serde_json::Value>,
-) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
-    unimplemented!();
+    /// Given the set of credentials, create a unsigned JSON-LD Presentation of those credentials.
+    /// In order to become a Verifiable Presentation, a data integrity proof must be created for the presentation and appended to the JSON-LD document.
+    pub fn create_presentation(
+        _creds: Vec<serde_json::Value>,
+    ) -> Result<serde_json::Value, Box<dyn std::error::Error>> {}
+        // unimplemented!();
```
LuisOsta, line 50: This should be uncommented it out

registry-resolver/Cargo.toml
```diff
@@ -11,6 +11,9 @@ prost-types = "0.10.1"
 tonic = { version = "0.7.1", features = ["compression"] }
 ssi = {path = "../core"}
 serde_json = "1.0.81"
+signature = "1.5.0"
+map = "0.0.0"
```
LuisOsta, line 15: Why were these dependencies added?

core/src/credential.rs
```diff
@@ -0,0 +1,147 @@
...
+    verifiable_credential:VerifiableCredential,
+    proof: *mut CredentialProof,
+}
+
+pub struct CredentialManager {
+    credential_subject: HashMap<&str, &str>,
+    context: VCContext,
+}
+
+trait VC {
```
LuisOsta, line 47: I don't think I quite explained myself, these trait functions should be moved to `DocumentBuilder`. As that's the public facing trait users will actually be easily able to override

core/src/credential.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn getSubject(&self, key: &str) -> Result<String, Error> {
```
LuisOsta, line 48: I think the param here should be called `cred_type`

core/src/credential.rs
```diff
@@ -0,0 +1,147 @@
...
+	subject: CredentialSubject,
+}
+
+pub struct IssuedCredential <'a> {
+    verifiable_credential:VerifiableCredential,
+    proof: *mut CredentialProof,
+}
+
+pub struct CredentialManager {
+    credential_subject: HashMap<&str, &str>,
```
LuisOsta, line 43: Shouldn't the return value here be `json_serde::Value`?

core/src/credential.rs
```diff
@@ -0,0 +1,147 @@
...
+use serde_json::Value;
+
+// cred_subject is a generic that implements trait X
+// trait X allows us to encode that object into JSON-LD
+// We provide types that implement trait X for the cred types that we support
+// Users can also user their own types that implement trait X if they need a different structure
+// --- 
+// Default context and Cred types are defaulted but can be redefined 
+
+type VCContext = [&'static str];
```
LuisOsta, line 16: Why make this a custom type? And also I think we should rename this type (if we're gonna keep it) into something more generic (since its for both VCs and VPs)

core/src/credential.rs
```diff
@@ -0,0 +1,147 @@
...
+
+pub(crate) const JSON_LD_CONTEXT:  VCContext = ["https://www.w3.org/2018/credentials/v1", "https://www.w3.org/2018/credentials/examples/v1"];
+
+pub const CRED_TYPE_PERMANENT_RESIDENT_CARD: str = "PermanentResidentCard";
```
LuisOsta, line 20: How about we remove the `CRED_TYPE` prefix and create a custom type called `CredentialType` or something like that?

core/src/credential.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+trait VC {
+    fn getSubject(&self, key: &str) -> Result<String, Error> {
+        match self.credential_subject.get(key) {
+            Some(&value) => Ok(value.to_string()),
+            None => Err(Error::new(ErrorKind::Other, "Key not found")),
+        } 
+    }
+    fn getContext(&self) -> VCContext {
```
LuisOsta, line 54: I think what we should expect here is just a basic string array. I think requiring them to use our types isn't necessary here

core/src/credential.rs
```diff
@@ -0,0 +1,147 @@
...
+                    ]
+                 }"###),
+            ]);
+        return Self { 
+            credential_subject: cred_subject_template,
+            context: JSON_LD_CONTEXT,
+         }
+    }
+
+    pub fn initVerifiableCredential (&self, cred_type: String, cred_subject: serde_json::Value, issuer: &'a str, id: &'a str) -> Result(VerifiableCredential, Error) {
```
LuisOsta, line 103: This function should receive the subject and context via parameters. The `DocumentBuilder` will use the override functions. I think our actual credential creation functionality should be as "pure" and stateless as possible in order to make it easier to test and debug

LuisOsta: So we separate the "data collection" i.e. defining what pieces we'll use to create the VC. And the actual process of creation itself.

Gathering -> Construction

core/src/credential.rs
LuisOsta: We are also missing a trait function called `getCredentialTypes`. Which returns the list of credential types (as strings) that we support

LuisOsta:
Oh also this is the JSON-Ld library we'll be using - https://docs.rs/json-ld/latest/json_ld/index.html. Might be worth a look @sKudryashov

core/src/credential.rs
sKudryashov: Agree, it can be enum based type, it is enum by nature

core/src/credential.rs
sKudryashov: Not sure, because so far it array of contexts. But what if we change this array and decide to bind it to a certain credential type? I remember it was discussed, so far I've implemented the simplest option, but it can be a subject for a change, that's why it is a custom type. We can change the underlying type but VCContext in all functions signatures will remain the same. Just for flexibility

LuisOsta: Alright that's fair I can see the value of that

core/src/credential.rs
LuisOsta: Yeah as long as we also get the string format without issue then it'll work out 👍🏽

core/src/credential.rs
LuisOsta: The custom type thing has been addressed, though the rename would still be valuable

core/src/credential.rs
sKudryashov: The idea is to maintain string in template and convert to `json_serde::Value` later, maybe we need to convert it to interim state to change some fields.

sKudryashov: As far as i remember there was a discussion with @fkim7 where we were discussing an opportunity to have some kind of credential_subject templates dependent on the credential types, correct me if I'm wrong. The implementation allows end-users to redefine that via trait but by default, we provide two credential types and two corresponding credential subjects.

sKudryashov: Here is a link for the reference (see above also) https://github.com/knox-networks/ssi/pull/5#discussion_r908461052

core/src/credential.rs
sKudryashov: Functions like create_credential and create_presentation we were talking about moved to `DocumentBuilder` trait indeed. VC trait and its implementation for Credential manager operates one level below. It allows you to use default `CredentialContext,` `CredTypes` and corresponding `CredentialSubjects` but gives an opportunity to overlap them by implementing your own trait, keeping the same `VerifiableCredential` struct. Unlike re-Implementing `DocumentBuilder` that let you use any type instead of `VerifiableCredential`.

core/src/credential.rs
LuisOsta: I think we had decided against it. The end user might use templates but we should expect the full credential subject to be passed down to the `create_credential` function. I think the templating approach is too complex/cumbersome at this level of abstraction

core/src/credential.rs
sKudryashov: I've tried to figure out the best possible structure for this, but const approach seems to be more suitable here for a couple of reasons:
- neither enum nor Struct is iterable which still puts on us the responsibility to maintain the array of available credentials manually, the same as with constants, so this approach doesn't really give us anything 
- both enum and struct require initialization in order for these two fields to be ready for usage, moreover there is no clear semantical definition that this is constant and can't be changed
- with const approach user sees predefined values for this library without any initialization and can use these constants in the method calls right away, like:
`ssi.creade_credential(ssi.CRED_TYPE_CARD, param1, param2)`

core/src/credential.rs
LuisOsta: @sKudryashov When the end-user would override `DocumentBuilder` they can only override what they need. I just don't the see the value of the extra abstraction with the `CredentialManager`. If they want to do a full override they can still use the struct that we defined for that override

core/src/credential.rs
```diff
@@ -0,0 +1,143 @@
...
+                (CRED_TYPE_PERMANENT_RESIDENT_CARD, r###"{
+                    "@id": "https://w3id.org/citizenship#PermanentResidentCard",
+                    "@context": {
+                      "@version": 1.1,
+                      "@protected": true,
+                      "id": "@id",
+                      "type": "[
+                        "PermanentResidentCard"
+                     ]",
+                      "description": "http://schema.org/description",
```
fkim7, line 92: Please note description, identifier, name are outside of the credential subject and belong in a level up.  Also please check these values- they have to be passed in by the user, and are not these predefined string values
  "id": "https://issuer.oidp.uscis.gov/credentials/83627465",
  "type": ["VerifiableCredential", "PermanentResidentCard"],
  "issuer": "did:example:28394728934792387",
  "identifier": "83627465",
  "name": "Permanent Resident Card",
  "description": "Government of Example Permanent Resident Card.",
  "issuanceDate": "2019-12-03T12:19:52Z",
  "expirationDate": "2029-12-03T12:19:52Z",
  "credentialSubject": {
    "id": "did:example:b34ca6cd37bbf23",
    "type": ["PermanentResident", "Person"],
    "givenName": "JOHN",
    "familyName": "SMITH",
    "gender": "Male",
    "image": "data:image/png;base64,iVBORw0KGgo...kJggg==",
    "residentSince": "2015-01-01",
    "lprCategory": "C09",
    "lprNumber": "999-999-999",
    "commuterClassification": "C1",
    "birthCountry": "Bahamas",
    "birthDate": "1958-07-17"
  },

core/src/credential.rs
LuisOsta: My concern is that the extra abstraction seems to not add value while adding complexity. I want to make sure we take as minimal of an approach as possible since this functionality will be fairly low level. And I think having the `DocumentBuilder` be a trait that the user can leverage and optionally override to whatever measure they need is the approach that best balances ease-of-use with complexity

registry-resolver/Cargo.toml
```diff
@@ -11,6 +11,7 @@ prost-types = "0.10.1"
 tonic = { version = "0.7.1", features = ["compression"] }
 ssi = {path = "../core"}
 serde_json = "1.0.81"
+serde = { version = "1.0", features = ["derive"] }
```
LuisOsta, line 14: Why did serde have to get added here?

core/src/lib.rs
```diff
@@ -14,23 +23,32 @@ pub trait DIDResolver {
...
+    /// In order to become a Verifiable Credential, a data integrity proof must be created for the credential and appended to the JSON-LD document.
+    /// this is the default implementation of the `create` method. The `create` method can be overridden to create a custom credential.
+    pub fn create_credential(
+        &self,
+        cred_type: String, 
+        cred_subject: HashMap<String, Value>,
+        property_set: HashMap<String, Value>,
+        id: &str
+    ) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
+        let cm = CredentialManager::new();
```
LuisOsta, line 37: Still think we should do away with the `CredentialManager`

LuisOsta (changes requested):

core/src/credential.rs
```diff
@@ -0,0 +1,82 @@
...
+    #[serde(flatten)]
+    pub property_set: HashMap<String, Value>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Credential <'a> {
+    #[serde(flatten)]
+    verifiable_credential:VerifiableCredential <'a>,
+    #[serde(flatten)]
+    proof: CredentialProof <'a>,
```
LuisOsta, line 35: A credential shouldn't have a proof. And also it shouldn't be flattened

core/src/credential.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Serialize, Deserialize, Clone)]
+struct CredentialSubject {
+    #[serde(flatten)]
+    pub property_set: HashMap<String, Value>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Credential <'a> {
+    #[serde(flatten)]
+    verifiable_credential:VerifiableCredential <'a>,
```
LuisOsta, line 33: I don't think this is a good approach from a strongly typed situation. I think it would be too difficult to utilize. All of the properties should be defined here as normally. I think we should have three types

- Credential - All of the standard properties we defined (non-flattened) and then one property called "custom_properties" or something like that which does get flattened
- IntegrityProof - Same as what we currently have as CredentialProof. Just a more generic name that applies to both VCs and VPs
- Verifiable Credential - The union of the Credential and IntegrityProof - https://doc.rust-lang.org/reference/items/unions.html.

core/src/credential.rs
```diff
@@ -0,0 +1,82 @@
...
+
+type VerificationContext = [&'static str];
+
+pub const CONTEXT_CREDENTIALS:  VerificationContext = ["https://www.w3.org/2018/credentials/v1", "https://www.w3.org/2018/credentials/examples/v1"];
+
+pub const CRED_TYPE_PERMANENT_RESIDENT_CARD: str = "PermanentResidentCard";
+pub const CRED_TYPE_BANK_CARD: str = "BankCard";
+
+#[derive(Serialize, Deserialize, Clone)]
+struct CredentialSubject {
```
LuisOsta, line 25: We should add the `id` property here

core/src/lib.rs
```diff
@@ -14,23 +23,36 @@ pub trait DIDResolver {
...
-    unimplemented!();
-}
+pub trait DocumentBuilder {
+    /// Given the credential type and the credential subject information, create a unissued JSON-LD credential.
+    /// In order to become a Verifiable Credential, a data integrity proof must be created for the credential and appended to the JSON-LD document.
+    /// this is the default implementation of the `create` method. The `create` method can be overridden to create a custom credential.
+    pub fn create_credential(
+        &self,
+        cred_type: String, 
+        cred_subject: HashMap<String, Value>,
```
LuisOsta, line 33: Why not require the library consumer to pass something of type `CredentialSubject` here?

LuisOsta (changes requested):

core/src/lib.rs
```diff
@@ -14,23 +23,36 @@ pub trait DIDResolver {
...
+        property_set: HashMap<String, Value>,
+        id: &str
+    ) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
```
LuisOsta, line 36: Also here we should probably just return `vc` instead of the `serde_json::Value` serialization. We should let the user decide if they want to serialize it. 

Since we're so level we would want to make as few decisions for the user as possible and serialization is a big enough deal to be hesitant to do so. Also the user may need to do some things with the VC afterwards and JSON is harder to deal with than an actual struct

core/src/lib.rs
sKudryashov: It's just for consistency. The same way we are handling property_set that will be unpacked in VerifiableCredential struct, the same way we are accepting key-value to be unpacked in CredentialSubject struct. In both cases we are accepting key-value sets to unpack them later in corresponding structures

core/src/lib.rs
sKudryashov: Yes, completely agree, it is always simpler to handle objects. Our first interface was indeed JSON but know it seems to be evident that objects are more suitable here. Moreover later we can accept this object as an input for subsequent calls

core/src/credential.rs
sKudryashov: Sorry didn't get the idea of union usage. Moreover, to me it looks like we or lib clients have to use unsafe mode since: 

> Consequently, all reads of union fields have to be placed in unsafe blocks: https://doc.rust-lang.org/reference/items/unions.html

LuisOsta: We definitely want to avoid using unsafe mode if at all possible

LuisOsta: My bigger issue is with nesting the VerifiableCredential type inside of the Credential type

LuisOsta: If anything it should be the reverse but I'd rather not do it at all, I think its unnecessarily confusing and seems like a bit of a hack around the limitation of union types

core/src/credential.rs
```diff
@@ -0,0 +1,93 @@
...
+#[serde(bound(deserialize = "'de: 'static"))]
+pub struct Credential {
+    #[serde(flatten)]
+    verifiable_credential:VerifiableCredential,
+    proof: IntegrityProof,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(bound(deserialize = "'de: 'static"))]
+pub struct VerifiableCredential {
```
mkatychev, line 44: You can use a [`#[serde(rename_all = "camelCase")]`](https://serde.rs/attr-rename.html) header for the struct whilst keeping explicit fields with their own renames (such as id). Here's an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cbedd42e668e35691152b90bec704e65)

core/src/credential.rs
```diff
@@ -0,0 +1,93 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(bound(deserialize = "'de: 'static"))]
+pub struct Credential {
+    #[serde(flatten)]
+    verifiable_credential:VerifiableCredential,
+    proof: IntegrityProof,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(bound(deserialize = "'de: 'static"))]
```
mkatychev, line 43: Looks like we can't suppress this warning yet:
https://github.com/rust-lang/rust/issues/96956

Once that's merged into main we can add `#[allow(needless_lifetimes)]` here
For now, let's add a comment linking to this post: 
https://users.rust-lang.org/t/restrict-generic-type-to-static-lifetime/16915/4 

The borrow checker currently isn't smart enough to reason around `'de: 'static` bounds.

@LuisOsta @sKudryashov LMK if you have any questions about the two links above

core/src/credential.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct VerifiableCredential {
+    #[serde(rename = "@context")]
+    context:  VerificationContext,
+    #[serde(rename = "@id")]
+    id: String,
+    #[serde(rename = "type")]
+    cred_type: String,
```
mkatychev, line 50: if you wanted to keep a field name of type, you _could_ call it `r#type: String,` it will look weird but this way you can explicitly have the field named `self.r#type` if you'd like

mkatychev (approved):
LGTM
@sKudryashov  we can follow up with suggested changes in a separate PR. The comment section shouldn't hit triple digits 😓 

@LuisOsta  any concerns about merging as is?

core/Cargo.toml
```diff
@@ -6,6 +6,9 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+assert-json-diff = "2.0.2"
```
LuisOsta, line 9: this should be a dev dependency

core/src/credential.rs
```diff
@@ -0,0 +1,93 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug)]
+struct CredentialSubject {
+    id : String,
+    #[serde(flatten)]
+    pub property_set: HashMap<String, Value>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(bound(deserialize = "'de: 'static"))]
+pub struct Credential {
```
LuisOsta, line 36: Same issue as mentioned before with this struct and its relationship to `VerifiableCredential`.

A few things:
1. A "Credential" will not have a proof as its not issued
2. I think the Credential should have its own properties and we shouldn't tie them in this way.

core/src/lib.rs
```diff
@@ -22,23 +28,37 @@ pub trait DIDResolver {
...
+    /// In order to become a Verifiable Credential, a data integrity proof must be created for the credential and appended to the JSON-LD document.
+    /// this is the default implementation of the `create` method. The `create` method can be overridden to create a custom credential.
+    fn create_credential(
+        &self,
+        cred_type: String, 
+        cred_subject: HashMap<String, Value>,
+        property_set: HashMap<String, Value>,
+        id: &str
+    ) -> Result<VerifiableCredential, Box<dyn std::error::Error>> {
+        let vc = VerifiableCredential::new(CONTEXT_CREDENTIALS,
```
LuisOsta, line 42: This function should return a Credential not a VerifiableCredential. In order to create a VC we would need to create a data integrity proof which is the responsibility of the `create_data_integrity_proof`. In a later PR we can maybe investigate how to ease that process.

LuisOsta (changes requested):
@mkatychev @sKudryashov There's a couple of issues but apart from them I'm happy to approve this PR

core/src/credential.rs
sKudryashov: What if we return Credential, and will have a separate type for the signed one, like "VerifiedCredential"?

sKudryashov: And the relationship between them will be like VerifiedCredential contains Credential and all Credential fields will be flattened in VerifiedCredential? The idea behind it is that there is no point in maintaining a separate structure for the unsigned and signed credential, since they are essentially the same.

LuisOsta: Yeah a VerifiableCredential = Credential + Proof. I agree that it's not great to have to support both. But using the serde flattening feels like a hack around the limitations of the union type. I don't know what the Rust standard is for these situations. @mkatychev?

Rust type unions require unsafe reading which is why I think that not using them is a good idea. But I'm curious as to what the recommended Rust solution is here. Apart from maintaining two structs which is definitely unideal but seems to me better than relying on serde to flatten the definitions and hence doing a runtime union

LuisOsta: @sKudryashov @mkatychev If there is no way of doing something like a Union then let's just keep it simple and keep the flattening. So the only changes I'd need to see is:

1. create_credential should return a `Credential`
2. The relationship should be flipped s.t a VerifiableCredential contains the Credential and a Proof

mkatychev: @LuisOsta  unless this PR needs to be closed to address your concerns, let's create a ticket  or open a discussion and merge this one. There's too many disparate discussions in this PR and too much to discuss in general.

LuisOsta: @mkatychev I agree that this has gone too long but points 1 and 2 raised in my last comment are easy changes and fundamental to this PR (i.e. it'll be non functional if merged)

mkatychev: Yeah they both sound like reasonable changes. I'm not sure if `VerfiableCredential` is causal to `Credential` or the other way around, I suppose that conclusion would guide me towards 1 or 2.

core/src/credential.rs
sKudryashov: Thanks, this is a great trick, since the `type` word is reserved

sKudryashov: I'll keep it as is so far but thank you for introducing it, it is definitely good for the similar situations in the future

core/src/credential.rs
```diff
@@ -0,0 +1,93 @@
...
+    id : String,
+    #[serde(flatten)]
+    pub property_set: HashMap<String, Value>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(bound(deserialize = "'de: 'static"))]
+pub struct VerifiableCredential {
+    #[serde(flatten)]
+    verifiable_credential:Credential,
```
LuisOsta, line 38: This should just be `credential`

LuisOsta (approved):
One minor nit remaining but looks good to me overall

PR #6, opened 2022-06-22
Refactor Repository To Support DID Resolution & Code Generation

LuisOsta:
1. Refactored to Rust monorepo
2. Integrated Buf code generation into Registry Resolver


Cargo.toml
```diff
@@ -1,10 +1,5 @@
...
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
-[dependencies]
-serde_json = "1.0.81"
-signature = "1.5.0"
+[workspace]
+members = [
+	"core",
+	"registry-resolver"
+]
```
sKudryashov, line 5: NIT: Newline pls

sKudryashov (approved):
Fix NIT please

Ryanmtate (approved):
Looks great-- just a minor suggestion on the use of `#[path]` over `include!`, but can be modified later if you find one to be more useful than the other.

registry-resolver/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+pub mod registry {
+    include!("gen/registry_api.v1.rs");
```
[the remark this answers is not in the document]
LuisOsta: I didn't run into any issues but I'll migrate it over if that's the better approach

registry-resolver/src/lib.rs
```diff
@@ -1,6 +1,5 @@
-pub mod registry {
-    include!("gen/registry_api.v1.rs");
-}
+#[path = "gen/registry_api.v1.rs"]
+pub mod registry {}
```
Ryanmtate, line 2: okay to drop the `{}`

`pub mod registry` should work by itself.

PR #16, opened 2022-06-27
10 Implement `create_verification_method` and `get_method` in `RegistryResolver`

LuisOsta:


LuisOsta:
Still need to add docs

registry-resolver/src/lib.rs
```diff
@@ -15,6 +17,10 @@ impl RegistryResolver {
 }
 #[async_trait::async_trait]
 impl ssi::DIDResolver for RegistryResolver {
+    fn get_method() -> String {
```
mkatychev, line 20: this can be [`const fn`](https://doc.rust-lang.org/std/keyword.const.html#compile-time-evaluable-functions)

registry-resolver/src/lib.rs
```diff
@@ -4,6 +4,8 @@ use std::collections::HashMap;
 #[path = "gen/registry_api.v1.rs"]
 pub mod registry;
 
+const DID_METHOD: &str = "knox";
```
sKudryashov, line 7: Maybe `&'static str` ?

sKudryashov (changes requested):

registry-resolver/src/lib.rs
LuisOsta: @mkatychev I don't think a function in a trait can be a const, at least I get an error when I try

mkatychev: @LuisOsta yes you'll have to make it a standalone function and call it from within the method.

LuisOsta: I don't understand, what's the value of doing that?

LuisOsta: Interesting, if we get the trait method `get_method` and just call a const function, why is it cheaper? Like how does that work?

LuisOsta: I've implemented the helper function @mkatychev , though some more context as to how this pattern actually works would be great

sKudryashov (approved):

PR #21, opened 2022-06-30
14 Implement Unit Tests For `create` in `RegistryResolver`

LuisOsta:


LuisOsta:
I'm aware that y'all may not be able to review right now. But just a quick summary:

- Utilized `mockall` to implement mocking of the registry client (which abstracts the network layer
- Refactored the library to create a new `RegistryClient` which is an extremely thin network layer separating the RegistryResolver from the network communications
- Utilized `rstest` to implement parametrized cases for the tests. Then utilized `async-std` for the async runtime.

Now that this is done, we'll have a good foundation to work from for all other tests

LuisOsta:
Also do y'all know if there is a way to separate the tests from the implementation code? (at least like separate files)

sKudryashov (approved):
LGTM

registry-resolver/src/lib.rs
```diff
@@ -59,12 +46,85 @@ impl ssi::DIDResolver for RegistryResolver {
...
-    fn test_create() -> Result<(), String> {
-        assert!(false);
-        Ok(())
+    use crate::{
+        registry_client::registry::CreateResponse, registry_client::MockRegistryClient,
+        RegistryResolver,
+    };
+
+    fn create_did_doc(did: String) -> serde_json::Value {
+        return serde_json::json!({
```
mkatychev, line 57: should we store this in a test json file and do an fs read?

LuisOsta: @mkatychev What's the benefit of that? I feel like storing it in a separate file reduces the context of it which increases the likelihood of misuse

registry-resolver/Cargo.toml
```diff
@@ -14,3 +14,8 @@ serde_json = "1.0.81"
 pbjson = "0.3"
 pbjson-types = "0.3"
 async-trait = "0.1.53"
+mockall = "0.11.1"
+
+[dev-dependencies]
+rstest = "0.15.0"
+async-std = { version = "1.5", features = ["attributes"] }
```
mkatychev, line 21: we decided on tokio for monetae runtime, would tokio work here as well en lieu of async-std?

LuisOsta: `rstest` works more easily with `async-std` which is why I migrated to it. Do you prefer tokio over async-std here? If so why?

LuisOsta: Here's the link to the relevant docs - https://docs.rs/rstest/latest/rstest/attr.rstest.html#async

registry-resolver/src/lib.rs
mkatychev: that's a good point I suppose I was thinking of it in terms of reusable test dummy data such as the directory [monetae/core has](https://github.com/monetae/core/tree/develop/integration/data). I see your point about cognitive load.

registry-resolver/Cargo.toml
mkatychev: that's reason enough, can we add the hyperlink as a comment above line 21?

LuisOsta: Done

mkatychev (approved):
🚀

PR #22, opened 2022-07-07
15 Implement Unit Tests For Read `RegistryResolver`

LuisOsta:


core/src/error.rs
```diff
@@ -0,0 +1,34 @@
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum ErrorKind {
+    DocumentNotFound,
+    InvalidData,
+    Uncategorized,
+    NetworkFailure,
+}
+
+impl std::fmt::Display for ErrorKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
sKudryashov, line 10: Can we move std::fmt:: to a "use" definition?

sKudryashov (approved):
NIT picking, consider moving std::fmt:: into use definition to avoid repetitiveness, in general LGTM

Issue #23, opened 2022-07-07
Improve The Error Handling & Support


LuisOsta:
- https://docs.rs/eyre/latest/eyre/ 
- https://docs.rs/thiserror/latest/thiserror/

LuisOsta:
We should remove our heavy usage of `map_err` in the `signature` library.

We should be able to leverage `thiserror` in order to heavily improve how we are handling errors.

PR #31, opened 2022-07-12
17 Implement Create Data Integrity Proof For Credentials & Presentations

LuisOsta:
So I had to pretty significantly refactor the way that we were going to initially handle signing in order for it to actually work with digital identity. I've created our own signature suite that will be flexible enough to actually handle the use cases that we have. So that end I've created a new `signature` package that the ssi core will utilize (as well as external users if needed).

I've tried to make it as compatible as possible with the way that monetae currently handles signing and verifying.

I have two open questions:

1. Is there still a point to using `signature` as the foundations for our traits? (Considering how much we've had to build on top of that?
2. How could I test the failure of `try_encoded_sign` in a semi-realistic way?


signature/src/suite.rs
```diff
@@ -0,0 +1,35 @@
...
+impl AsRef<[u8]> for Ed25519Signature {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl std::fmt::Display for VerificationRelation {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            VerificationRelation::AssertionMethod => write!(f, "assertionMethod"),
```
mkatychev, line 23: it's probably simpler to do an `From<VerificationRelation> for <&static str>` here and call `write!(f, self.into())`
or you could [get fancy with macros](https://github.com/mkatychev/tache/blob/master/src/lib.rs#L170-L235)!

core/src/lib.rs
```diff
@@ -12,11 +14,11 @@ pub trait DIDResolver {
     // Returns the DID Method that the DID Resolver is compatible with. Each resolver can only be compatible with one.
     fn get_method() -> &'static str;
     // Given a `did` and `key` it will construct the proper `verificationMethod` to use as part of the data integrity proof creation process.
-    fn create_verification_method(did: String, key_id: String) -> String {
+    fn create_verification_method(public_key: String, key_id: String) -> String {
         return format!(
             "did:{}:{}#{}",
```
mkatychev, line 19: These two variables _should_ be able to be templated:
```suggestion
            "did:{}:{public_key}#{key_id}",
```

signature/src/verifier.rs
```diff
@@ -0,0 +1,106 @@
...
+        sig: &Ed25519Signature,
+        relation: VerificationRelation,
+    ) -> Result<(), signature::Error> {
+        let sig_bytes: [u8; 64] = sig
+            .0
+            .as_slice()
+            .try_into()
+            .map_err(|_| signature::Error::new())?;
+        let sig = ed25519_zebra::Signature::from(sig_bytes);
+        match relation {
```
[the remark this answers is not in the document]
LuisOsta: @mkatychev Yes for now but this structure is necessary since each Verification Relation will have to have its own public key in the (upcoming) feature. So I wanted to build out the structure now to support it

signature/src/verifier.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    fn decode(&self, encoded_sig: String) -> Result<S, signature::Error>;
+}
+
+impl signature::Verifier<Ed25519Signature> for Ed25519DidVerifier {
+    fn verify(&self, msg: &[u8], sig: &Ed25519Signature) -> Result<(), signature::Error> {
+        let sig_bytes: [u8; 64] = sig.0.as_slice().try_into().unwrap();
+        self.public_key
+            .verify(&ed25519_zebra::Signature::from(sig_bytes), msg)
+            .map_err(|_| signature::Error::new())
```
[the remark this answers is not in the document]
LuisOsta: @mkatychev Both are from external crates so it doesn't seem like I can do that?

LuisOsta:
I think the best thing to do here is to remove our usage of the `signature` traits defined here - https://github.com/RustCrypto/traits/tree/master/signature. Do y'all agree? @sKudryashov @mkatychev @Ryanmtate . They are a bit of a pain to utilize and support because they're not really built for the kind of flexible use case we need for `ssi` (namely multiple keys and relational keys).

I was thinking that we just inline the interface that we need and remove the dependency altogether. Instead of trying to extend them and retrofit the use case to them

signature/src/verifier.rs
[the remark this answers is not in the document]
mkatychev: Isn't `signature::Error` in SSI?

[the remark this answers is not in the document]
LuisOsta: That's actually defined by https://github.com/RustCrypto/traits/tree/master/signature, we have a bit of a weird dependency with them which is why I proposed that we remove that altogether since its not necessary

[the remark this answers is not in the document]
mkatychev: I see, or you can wrap a signature::Error using `thiserror` that way you have control over your methods.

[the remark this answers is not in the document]
LuisOsta: I've removed signature, let me know what you think

signature/src/suite.rs
LuisOsta: What is the benefit of doing that? Wouldn't we still need to do the match set in that one?

LuisOsta:
So there are a few open questions (at least for me):

- Right now the test coverage I added was for the core library and not `signature`, that was because I didn't see much value in testing the `signature` library since it was basically just cryptography. And I felt like there was sufficient coverage from testing the higher level stuff (I may come to regret those words Idk). Do y'all agree
- Do you think I'm missing some coverage for `create_data_integrity_proof`?
- Any other areas you think could benefit from a refactor?

signature/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+ed25519-zebra = "3.0.0"
+multibase = "0.9.1"
+rand = "0.8.5"
+serde = { version = "1.0", features = ["derive"] }
```
sKudryashov, line 12: Newline pls

signature/src/error.rs
```diff
@@ -0,0 +1,66 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub enum Error {}
```
sKudryashov, line 4: NIT: I thing the newline has to be here

signature/src/signer.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        return Self {
+            private_key: sk,
+            public_key: ed25519_zebra::VerificationKey::from(&sk),
+        };
+    }
+}
+
+impl DIDSigner<Ed25519Signature> for Ed25519DidSigner {
+    fn try_sign(&self, data: &[u8]) -> Result<Ed25519Signature, SignatureError> {
```
sKudryashov, line 43: In some cases we have newline between methods, in some we don't, maybe we have to be consistent here

signature/src/suite.rs
```diff
@@ -0,0 +1,46 @@
+use crate::error::SignatureError;
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
sKudryashov, line 2: Newline above?

sKudryashov (changes requested):

core/src/credential.rs
```diff
@@ -56,29 +55,20 @@ pub struct Credential {
...
-    proof_type: String,
-    created: String,
-    verification_method: String,
-    proof_purpose: String,
-    proof_value: String,
-}
-
 impl Credential {
-    pub fn new (
+    pub fn new(
```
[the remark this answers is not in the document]
Ryanmtate: It's already done? Why does removal of a space matter? Just leave it. Needless nit.

signature/src/verifier.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn from(signer: &crate::signer::Ed25519DidSigner) -> Self {
+        Self {
+            public_key: signer.public_key,
+        }
+    }
+}
+
+impl DIDVerifier<Ed25519Signature> for Ed25519DidVerifier {
+    fn verify(&self, msg: &[u8], sig: &Ed25519Signature) -> Result<(), SignatureError> {
+        let sig_bytes: [u8; 64] = sig.0.as_slice().try_into().unwrap();
```
Ryanmtate, line 46: Careful of using .`unwrap()` here, do not want to core dump.

Ryanmtate: Maybe worth adding a `thiserror` enum to `SignatureError` or keeping a higher-level `Error` type that encapsulates all error cases:

```rust
#[error(transparent)]
TryFromSlice(#[from] std::array::TryFromSliceError),
```

then you use the `?` operand, e.g. `let sig_bytes: [u8; 64] = sig.0.as_slice().try_into()?`

signature/src/suite.rs
LuisOsta: Sounds good

signature/src/verifier.rs
mkatychev: @Ryanmtate  I don't think that a rust panic [core dumps by default](https://dev.to/jcarlosv/core-dump-for-rust-10nm). I may be wrong though.

signature/src/error.rs
```diff
@@ -0,0 +1,67 @@
...
+            .source
+            .as_ref()
+            .map(|source| source.as_ref() as &(dyn std::error::Error + 'static));
+    }
+}
+
+impl From<ed25519_zebra::Error> for SignatureError {
+    fn from(e: ed25519_zebra::Error) -> Self {
+        SignatureError {
+            message: String::from(e.to_string()),
```
sKudryashov, line 62: I think it has to be `message: (e.to_string()` since to_string() returns String

sKudryashov (changes requested):

signature/src/error.rs
LuisOsta: Done

signature/src/error.rs
```diff
@@ -0,0 +1,67 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub enum Error {}
```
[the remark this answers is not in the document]
LuisOsta: @Ryanmtate We most likely will, I just wanted to handle that as part of #23 so that we can keep the ball moving. This PR changes a lot so I wanted to close it as quick as possible

sKudryashov (approved):

PR #34, opened 2022-07-13
Create Cargo Linting Check Github Action

LuisOsta:


registry-resolver/src/registry_client.rs
```diff
@@ -1,6 +1,7 @@
 use mockall::*;
 use registry::registry_service_client::RegistryServiceClient;
 
+#[rustfmt::skip]
```
rtsuk, line 4: What's rustfmt doing here that you don't like?

rtsuk (approved):

registry-resolver/src/registry_client.rs
LuisOsta: That module is generated code that isn't following the coding standards. But that's generated code so we don't really have control over those details and it doesn't really matter

PR #37, opened 2022-07-16
30 implement create verifiable presentation

sKudryashov:


core/src/lib.rs
```diff
@@ -220,14 +186,136 @@ mod tests {
             .entry("birthDate".to_string())
             .or_insert(Value::String("1958-07-17".to_string()));
 
+        return (kv_body, kv_subject);
+    }
+
+    #[test]
+    fn test_create_credential() -> Result<(), String> {
```
sKudryashov, line 193: Guys, Is it possible to create subtests in Rust with closures, how it can be done in GoLang: https://github.com/knox-networks/gomono/pull/570 @mkatychev @LuisOsta ? Thus we could distinct VC creation test from VC creating proof and from Creating Verifiable presentation. Not it is implemented in one test, but maybe we can do subtests like in aforementioned GoLang PR.

sKudryashov, line 193: Guys, Is it possible to create subtests in Rust with closures, how it can be done in GoLang: https://github.com/knox-networks/gomono/pull/570 @mkatychev @LuisOsta @Ryanmtate ? Thus we could separate VC creation test from VC creating proof and from Creating Verifiable presentation. Not it is implemented in one test, but maybe we can do subtests like in example PR above?

sKudryashov, line 193: Guys, Is it possible to create subtests in Rust with closures, how it can be done in GoLang: https://github.com/knox-networks/gomono/pull/570 @mkatychev @LuisOsta ? Thus we could distinct VC creation test from VC creating proof and from Creating Verifiable presentation. Not it is implemented in one test, but maybe we can do subtests like in aforementioned GoLang PR.

Ryanmtate: Yes, this is possible; I built something like this for the integrations test harness: 

https://github.com/monetae/core/blob/develop/integration/src/lib.rs#L45-L53
https://github.com/monetae/core/blob/develop/integration/src/lib.rs#L140-L146
https://github.com/monetae/core/blob/develop/integration/src/issuer.rs#L229-L232
https://github.com/monetae/core/blob/develop/integration/bin/main.rs#L43-L67

I can help configure something similar

LuisOsta: 6t5Why do you need subtests to separate the testing of VCs and VPs? Why not just create a new test like:

```
#[test]
fn test_create_presentation() -> Result<(), String> {
...
}
```
?
And any shared logic can be pulled out into a standard function in the test module?

LuisOsta (changes requested):

core/src/lib.rs
sKudryashov: It seems the comments might get duplicated. Yes that's an option and it looks like it is a single option for now, with closure subtest it could be just a more concise way.

core/src/credential.rs
```diff
@@ -80,4 +94,45 @@ impl Credential {
     pub fn serialize(&self) -> Value {
         return serde_json::to_value(&self).unwrap();
     }
+
+    pub fn create_verifiable_credentials(
+        &self,
+        integrity_proof: crate::proof::DataIntegrityProof,
+    ) -> VerifiableCredential {
+        let vc = VerifiableCredential {
+            credential: self.to_owned(),
```
mkatychev, line 103: if we're taking ownership of self here then we should just use `self` instead of `&self`
```suggestion
    pub fn create_verifiable_credentials(
        self,
        integrity_proof: crate::proof::DataIntegrityProof,
    ) -> VerifiableCredential {
        let vc = VerifiableCredential {
            credential: self,
```

core/src/credential.rs
```diff
@@ -80,4 +94,45 @@ impl Credential {
...
+    pub context: VerificationContext,
+    #[serde(rename = "verifiableCredential")]
+    pub verifiable_credential: Vec<VerifiableCredential>,
+}
+
+impl Presentation {
+    pub fn new(context: VerificationContext, vc: Vec<VerifiableCredential>) -> Presentation {
+        Presentation {
+            context: context,
+            verifiable_credential: vc,
```
mkatychev, line 131:
```suggestion
    pub fn new(context: VerificationContext, verifiable_credential: Vec<VerifiableCredential>) -> Presentation {
        Presentation {
            context,
            verifiable_credential,
```

core/src/lib.rs
```diff
@@ -110,13 +108,57 @@ mod tests {
...
+            ("issuanceDate", "2019-12-03T12:19:52Z"),
+            ("expirationDate", "2029-12-03T12:19:52Z"),
+        ])
+        .into_iter()
+        .map(|(k, v)| (k.into(), v.into()))
+        .collect();
+
+        kv_body
+            .entry("type".to_string())
+            .or_insert(json!(["VerifiableCredential", "PermanentResidentCard"]));
```
mkatychev, line 132: you can use [`HashMap::insert`](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.insert) for this one. Should be cleaner than `.entry(...).or_insert(...)`

core/src/lib.rs
```diff
@@ -145,89 +187,101 @@ mod tests {
...
+                    .proof_value
+                    .clone(),
+            );
+        expect_presentation["verifiableCredential"][0]["proof"]["created"] =
+            serde_json::Value::String(
+                interim_presentation.verifiable_credential[0]
+                    .proof
+                    .created
+                    .clone(),
+            );
```
mkatychev, line 279: we could use `serde_json::to_value` here to convert all of `interim_presentation.verifiable_credential[0].proof`:
```suggestion
        let interim_proof =
            serde_json::to_value(interim_presentation.verifiable_credential[0].proof.clone())
                .unwrap();
        expect_presentation["verifiableCredential"][0]["proof"] = interim_proof;
```

mkatychev: Esp. since `DataIntegrityProof` already implements `serde::Serialize`
https://github.com/knox-networks/ssi/blob/9c62cff5ffb0cfce223211465b2f4ef9a112b046/core/src/proof.rs#L5-L12

core/src/lib.rs
```diff
@@ -145,89 +187,101 @@ mod tests {
...
+                .len()
+                > 0
+        );
+        assert!(
+            interim_presentation.verifiable_credential[0]
+                .proof
+                .proof_value
+                .len()
+                > 0
+        );
```
mkatychev, line 257:
```suggestion
        let interim_proof = &interim_presentation.verifiable_credential[0].proof;
        assert!(!interim_proof.verification_method.is_empty());
        assert!(!interim_proof.proof_value.is_empty());
```

mkatychev: We're checking to see the inverse of `interim_proof.proof_value.is_empty()` here so we prepend the `!`

core/src/lib.rs
```diff
@@ -145,89 +187,101 @@ mod tests {
...
+
+        assert!(proof.is_ok());
+
+        let verifiable_credential = credential.create_verifiable_credentials(proof.unwrap());
+        let credentials = vec![verifiable_credential];
+        let pre = to.create_presentation(credentials);
+
+        assert!(pre.is_ok());
+
+        let interim_presentation = pre.unwrap();
```
mkatychev, line 242: depending on your intent, there's a more verbose version of `.unwrap()` called [`.expect("message")`](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect) that allows you to be comment on your unwrapped and panicking error:
```suggestion
        let interim_presentation = to.create_presentation(credentials).expect("unable to create presentation from credentials");
```

core/src/lib.rs
sKudryashov: Implemented without subtests but with separate tests function. I think table-driven tests can be added later to this test suite.

mkatychev (approved):

core/src/lib.rs
```diff
@@ -145,89 +186,69 @@ mod tests {
...
         let vc = to.create_credential(
             crate::CRED_TYPE_PERMANENT_RESIDENT_CARD.to_string(),
             kv_subject,
             kv_body,
             "https://issuer.oidp.uscis.gov/credentials/83627465",
         );
+
         assert!(vc.is_ok());
-        assert_json_eq!(expect, vc.unwrap());
+        let credential = vc.unwrap();
```
mkatychev, line 199: Let's do the same `.expect("...")` pattern here rather than `assert!(this.is_ok()); let that = this.unwrap();`

mkatychev (approved):

core/src/lib.rs
```diff
@@ -145,89 +186,69 @@ mod tests {
...
+        assert_json_eq!(expect_credential, credential.serialize());
+
+        test_create_presentation(to, credential)
+    }
+
+    fn test_create_presentation(to: TestObj, credential: crate::Credential) -> Result<(), String> {
```
mkatychev, line 205: unless this is a unit test we probably shouldn't add a `test_` prefix to the function.

core/src/lib.rs
```diff
@@ -110,13 +108,56 @@ mod tests {
...
+        .map(|(k, v)| (k.into(), v.into()))
+        .collect();
+
+        kv_subject.insert("type".to_string(), json!(["PermanentResident", "Person"]));
+
+        return (kv_body, kv_subject);
+    }
+
+    #[test]
+    fn test_create_credential_and_presentation() -> Result<(), String> {
```
LuisOsta, line 158: Why are these combined? I think we should have two separate tests

core/src/lib.rs
LuisOsta, line 205: This should be its own test not part of the VC test flow

LuisOsta (changes requested):

core/src/lib.rs
```diff
@@ -52,9 +52,11 @@ pub trait DocumentBuilder {
...
     /// In order to become a Verifiable Presentation, a data integrity proof must be created for the presentation and appended to the JSON-LD document.
     fn create_presentation(
-        _creds: Vec<serde_json::Value>,
-    ) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
-        unimplemented!();
+        &self,
+        credentials: Vec<VerifiableCredential>,
+    ) -> Result<Presentation, Box<dyn std::error::Error>> {
+        let pre = Presentation::new(CONTEXT_CREDENTIALS, credentials);
+        Ok(pre)
```
LuisOsta, line 59: You can just do `Ok(Presentation::new(CONTEXT_CREDENTIALS, credentials))`

core/src/credential.rs
```diff
@@ -55,6 +55,20 @@ pub struct Credential {
     pub property_set: HashMap<String, Value>,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct IntegrityProof {
```
LuisOsta, line 59: We should replace `Identityproof` with the `DataIntegrityProof` object in the `proof.rs` file. Let's make sure we just have one source of truth on this

LuisOsta (changes requested):

LuisOsta (approved):

Issue #36, opened 2022-07-13
Implement `From<serde_json::Value>` Trait For VerifiableCredentials and Credential Struct


LuisOsta:

LuisOsta:
In order for our library to be generally useful. And for it to work with storage (to convert from a stored format to in memory). We need to be able to convert to `serde_json::Value` and back. This is made more complex due to the requirement to support optional properties.


Right now we support converting from our structs to `serde_json::Value`. But we do not support going from `serde_json::Value` to the structs.

sKudryashov:
Custom deserialization exlored, tried, and also explored custom deserealization/serialization by field. It is taken as an approach, the advantage is that you customize serialization/deserialization only for certain fields, the rest is being deserialized as usual.

PR #38, opened 2022-07-21
Verifiable Credential Manual deserialization

sKudryashov:
https://github.com/knox-networks/ssi/issues/36


core/src/credential.rs
```diff
@@ -66,7 +81,7 @@ impl Credential {
         let vc = Credential {
             context: context,
             id: id.to_string(),
-            cred_type: cred_type.to_string(),
+            cred_type: cred_type,
```
mkatychev, line 84:
```suggestion
            cred_type,
```

core/src/credential.rs
```diff
@@ -80,4 +95,62 @@ impl Credential {
     pub fn serialize(&self) -> Value {
         return serde_json::to_value(&self).unwrap();
     }
+    
+    pub fn deserialize (contents: String) -> Credential {
```
mkatychev, line 99: We should probably create a ticket to have these return results `Result<Credential, Error>`. JSON is one of the most error prone parts of the code and it's quite a hot path for panicking.

core/src/lib.rs
```diff
@@ -39,6 +39,7 @@ pub trait DocumentBuilder {
         property_set: HashMap<String, Value>,
         id: &str,
     ) -> Result<Credential, Box<dyn std::error::Error>> {
+        println!("\nCreating credential {}", cred_type);
```
mkatychev, line 42:
```suggestion
        println!("\nCreating credential {cred_type}");
```

core/src/lib.rs
```diff
@@ -116,6 +117,16 @@ mod tests {
         let mut kv_body: HashMap<String, Value> = HashMap::new();
         let mut kv_subject: HashMap<String, Value> = HashMap::new();
 
+        let type_rs = serde_json::to_value([
+            "VerifiableCredential".to_string(),
+            "PermanentResidentCard".to_string(),
+        ]);
```
mkatychev, line 123:
```suggestion
        let type_rs = json!(["VerifiableCredential", "PermanentResidentCard"]);
```

core/src/lib.rs
```diff
@@ -116,6 +117,16 @@ mod tests {
...
+        if type_rs.is_ok() {
+            kv_body
+                .entry("type".to_string())
+                .or_insert(type_rs.unwrap());
```
mkatychev, line 127: if we do `.unwrap()` it should be called as early as possible in the flow.

core/src/credential.rs
sKudryashov: I think, it can be done as a part of this ticket. In fact it is  already done.

core/src/credential.rs
```diff
@@ -39,22 +41,157 @@ pub struct VerifiableCredential {
...
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        let idate = DateTime::parse_from_rfc3339(&s);
+        if idate.is_ok() {
+            let sys_time = SystemTime::from(idate.unwrap());
+            return Ok(sys_time.clone());
+        } else {
+            println!("error: {}", idate.unwrap_err());
+            return Err(serde::de::Error::custom("Invalid date"))
```
mkatychev, line 190: we should reuse the same error message here

core/src/credential.rs
```diff
@@ -80,4 +88,147 @@ impl Credential {
...
         return serde_json::to_value(&self).unwrap();
     }
+    
+    pub fn deserialize (contents: String) -> Result<Credential, serde_json::Error> {
+        let cred = serde_json::from_str(&contents);
+        if cred.is_ok() {
+            return Ok(cred.unwrap());
+        } else {
+            return Err(cred.unwrap_err());
+        }
```
mkatychev, line 94:
```suggestion
        let cred = serde_json::from_str(&contents)?;
```

core/src/credential.rs
sKudryashov: Implemented with different data type

core/src/lib.rs
sKudryashov: Implemented with different data type

LuisOsta (approved):

core/src/credential.rs
```diff
@@ -39,22 +41,157 @@ pub struct VerifiableCredential {
...
+        date: &SystemTime,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let formatted = DateTime::<Utc>::from(*date).format(FORMAT);
+        let s = formatted.to_string();
+        let s = format!("{}", &s);
+        serializer.serialize_str(&s)
```
mkatychev: Does this work?
```suggestion
        let formatted = DateTime::<Utc>::from(*date).format(FORMAT);
        serializer.serialize_str(&formatted.to_string())
```

core/src/credential.rs
```diff
@@ -80,4 +84,134 @@ impl Credential {
...
+    {
+        let s = String::deserialize(deserializer)?;
+        let idate = DateTime::parse_from_rfc3339(&s);
+        if idate.is_ok() {
+            let sys_time = SystemTime::from(idate.unwrap());
+            return Ok(sys_time.clone());
+        } else {
+            return Err(serde::de::Error::custom(idate.unwrap_err()));
+        }
+    }
```
mkatychev, line 211: For the error that can also be greatly simplified:
```suggestion
pub fn deserialize<'de, D>(deserializer: D) -> Result<SystemTime, D::Error>
    where
        D: Deserializer<'de>,
    {
        use serde::de::Error;
        let s = String::deserialize(deserializer)?;
        let idate = DateTime::parse_from_rfc3339(&s)
            .map(SystemTime::from)
            .map_err(Error::custom)?;
        Ok(idate)
    }
```

I’m using `.map` and `.map_err` to pass in functions directly.
I don’t need a closure since `.map` expects a function of some kind  `SystemTime::from` as well as `|d| SystemTime::from(d)` are identical in that respect

core/src/credential.rs
```diff
@@ -80,4 +84,134 @@ impl Credential {
...
     }
+
+    pub fn deserialize(contents: String) -> Result<Credential, serde_json::Error> {
+        let cred = serde_json::from_str(&contents);
+        if cred.is_ok() {
+            return Ok(cred.unwrap());
+        } else {
+            return Err(cred.unwrap_err());
+        }
+    }
```
mkatychev, line 90:
```suggestion
pub fn deserialize(contents: String) -> Result<Credential, serde_json::Error> {
        serde_json::from_str(&contents)
    }
```

core/src/credential.rs
```diff
@@ -80,4 +84,129 @@ impl Credential {
...
+
+    pub fn serialize<S>(date: &SystemTime, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let formatted = DateTime::<Utc>::from(*date).format(FORMAT);
+        let s = formatted.to_string();
+        let s = format!("{}", &s);
+        serializer.serialize_str(&s)
+    }
```
mkatychev, line 197:
```suggestion
pub fn serialize<S>(date: &SystemTime, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        let utc = DateTime::<Utc>::from(*date).format(FORMAT);
        serializer.serialize_str(&format!("{utc}"))
    }
```

mkatychev: That should be a little simpler. I've also called the variable `utc` to clarify the conversion.

mkatychev (changes requested):

mkatychev (approved):

Issue #39, opened 2022-07-26
Add `KeyPair` trait & `SSIKeyPair` struct in `signature` package.


LuisOsta:
The trait should implement the following functions:
```
get_public_key(relation SignerRelation)
get_private_key(relation SignerRelation)
```

It should be purely informational. And then we should add functionality to our SSI Signer that converts a SSIKeyPair into a Signer

fkim7:
After talking to Sergey, it seems supporting a generic signer instead of the current working ed25519 signer will be difficult.  Want to revisit this after, especially as we currently have everything dependent on ed25519 and if we need to add support for more signature types ex Secp256k1, we'll need to update all the components including interbank to make it work.  Will go with the ed25519 signer for the short term.

PR #40, opened 2022-07-28
Trait keypair added

sKudryashov:


sKudryashov:
@LuisOsta just a first draft, to ensure it implements the idea

signature/src/keypair.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn get_public_key_enc(&self) -> String;
+    fn get_private_key_enc(&self) -> String;
+    fn get_public_key(&self) -> ed25519_zebra::VerificationKey;
+    fn get_private_key(&self) -> ed25519_zebra::SigningKey;
+}
+
+#[derive(Debug)]
+pub struct SSIKeyPair {
+    pub(crate) public_key_encoded : String,
+    private_key_encoded: String,
```
LuisOsta, line 11: Why not make the private key also directly public?

signature/src/keypair.rs
```diff
@@ -0,0 +1,46 @@
+pub trait KeyPair: core::fmt::Debug + Sized {
+    fn get_public_key_enc(&self) -> String;
+    fn get_private_key_enc(&self) -> String;
+    fn get_public_key(&self) -> ed25519_zebra::VerificationKey;
+    fn get_private_key(&self) -> ed25519_zebra::SigningKey;
+}
+
+#[derive(Debug)]
+pub struct SSIKeyPair {
```
LuisOsta, line 9: I think we should call this Ed25519KeyPair right? And the trait `SSIKeyPair`?

signature/src/keypair.rs
```diff
@@ -0,0 +1,46 @@
...
+    // signature_type: String, 
+    private_key: ed25519_zebra::SigningKey,
+    pub(crate) public_key: ed25519_zebra::VerificationKey,
+}
+
+impl SSIKeyPair {
+    fn new() -> Self {
+        let sk = ed25519_zebra::SigningKey::new(rand::thread_rng());
+        let vk = ed25519_zebra::VerificationKey::from(&sk);
+        let encoded_pk = multibase::encode(multibase::Base::Base58Btc, vk);
```
LuisOsta, line 21: Why not only store the raw version and we can encode on demand?

signature/src/keypair.rs
```diff
@@ -0,0 +1,46 @@
+pub trait KeyPair: core::fmt::Debug + Sized {
+    fn get_public_key_enc(&self) -> String;
```
LuisOsta, line 2: I think we should just name it `get_encoded_public_key`. Since `enc` could mean both encrypted or encoded in this context

LuisOsta (changes requested):
I have some small issues but your overall direction is correct

core/src/lib.rs
```diff
@@ -57,6 +57,14 @@ pub trait DocumentBuilder {
     ) -> Result<Presentation, Box<dyn std::error::Error>> {
         Ok(Presentation::new(CONTEXT_CREDENTIALS, credentials))
     }
+
+    fn restore_identity(
+        alias: String,
```
LuisOsta, line 62: The alias should not be necessary in this context? The alias in the wallet cli was necessary because of storage reasons

LuisOsta (changes requested):

signature/src/keypair.rs
sKudryashov: Yep, we can do that, although it is implemented as described in the ticket: https://github.com/knox-networks/ssi/issues/39 , but i can rename, np

core/src/lib.rs
sKudryashov: And the idea here is that neither in Creation Identity nor in Restore Identity we won't use storage, unlike walled-cmd in gomono, right? It is the responsibility of the caller, so I'm not counting on storage here

LuisOsta: Yeah exactly

signature/src/keypair.rs
LuisOsta: If you prefer this naming convention we can keep it, I just hadn't fully realized what I had said in the issue lol

signature/src/signer.rs
```diff
@@ -23,6 +24,7 @@ where
     fn get_proof_type(&self) -> String;
     fn get_verification_method(&self, relation: VerificationRelation) -> String;
     fn encode(&self, sig: S) -> String;
+    fn from(kp: T) -> Self;
```
LuisOsta, line 27: Can we require that the trait implement the `From` trait?

LuisOsta: Or is that not possible

signature/src/signer.rs
```diff
@@ -41,7 +43,7 @@ impl Ed25519DidSigner {
     }
 }
 
-impl DIDSigner<Ed25519Signature> for Ed25519DidSigner {
+impl DIDSigner<Ed25519Signature, crate::keypair::SSIKeyPair> for Ed25519DidSigner {
```
LuisOsta, line 46: Do you think its a problem that we have both a signature generic and a keypair generic? Do you think that'd cause problems?

signature/src/keypair.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub(crate) private_key: ed25519_zebra::SigningKey,
+    pub(crate) public_key: ed25519_zebra::VerificationKey,
+}
+
+impl SSIKeyPair {
+    fn new() -> Self {
+        let sk = ed25519_zebra::SigningKey::new(rand::thread_rng());
+        let vk = ed25519_zebra::VerificationKey::from(&sk);
+        return Self {
+            public_key_encoded: None,
```
LuisOsta, line 29: Why not remove these properties?

signature/src/keypair.rs
```diff
@@ -0,0 +1,64 @@
...
+            Some(ref pk) => pk.clone(),
+            None => {
+                let encoded_pk = multibase::encode(multibase::Base::Base58Btc, self.public_key);
+                self.public_key_encoded = Some(encoded_pk.clone());
+                return encoded_pk;
+            }
+        }
+    }
+    fn get_private_key_encoded(&mut self) -> String {
+        match self.private_key_encoded {
```
LuisOsta, line 49: What I actually meant is just to return the result of running:
```
multibase::encode(multibase::Base::Base58Btc, self.public_key)
```
Do you think we shouldn't do that?

LuisOsta: I generally want to avoid state mutations, so I'd just prefer storing the encoded version outright vs doing mutations not in the `new` function

LuisOsta (changes requested):

signature/src/keypair.rs
sKudryashov: Lazy-load implemented, once someone calls get_public_key_encoded, it converts the key to a string and stores it in the property.

signature/src/keypair.rs
sKudryashov: Yes, it makes sense, since this is a library, it can be stateless and put this responsibility to a caller

signature/src/keypair.rs
```diff
@@ -0,0 +1,188 @@
...
+                    DID_PREFIX,
+                    multibase::encode(multibase::Base::Base58Btc, self.master_public_key),
+                    multibase::encode(
+                        multibase::Base::Base58Btc,
+                        self.capability_delegation_public_key
+                    )
+                )
+            }
+        }
+    }
```
mkatychev, line 163: This can be simplified by assigning the match arm to a value:
```suggestion
pub fn get_verification_method(self, relation: crate::suite::VerificationRelation) -> String {
        let verification_key = match relation {
            crate::suite::VerificationRelation::AssertionMethod => self.assertion_method_public_key,
            crate::suite::VerificationRelation::Authentication => self.authetication_public_key,
            crate::suite::VerificationRelation::CapabilityInvocation => {
                self.capability_invocation_public_key
            }
            crate::suite::VerificationRelation::CapabilityDelegation => {
                self.capability_delegation_public_key
            }
        };
        format!(
            "{DID_PREFIX}{}#{}",
            multibase::encode(multibase::Base::Base58Btc, self.master_public_key),
            multibase::encode(multibase::Base::Base58Btc, verification_key)
        )
    }
```

core/Cargo.toml
```diff
@@ -12,8 +12,12 @@ serde_json = "1.0.81"
 sha2 = "0.10.2"
 signature = {path = "../signature"}
 serde = { version = "1.0", features = ["derive"] }
+mockall = "0.11.1"
 
 [dev-dependencies]
 rstest = "0.15.0"
 map = "0.0.0"
 assert-json-diff = "2.0.2"
+mockall = "0.11.1"
```
LuisOsta, line 21: Why is this both a general and a dev dependency?

core/src/lib.rs
```diff
@@ -1,15 +1,22 @@
+#![allow(unused_variables)]
+#![allow(dead_code)]
 mod credential;
-
 use credential::*;
 use serde_json::{self, Value};
+use signature::keypair::Ed25519SSIKeyPair;
```
LuisOsta, line 6: We should generally refrain from `use` and instead utilize things through the appropriate namespace i.e `signature::keypair::Ed25519....`

core/src/lib.rs
```diff
@@ -102,9 +101,10 @@ mod tests {
 
     impl TestObj {
```
LuisOsta, line 102: What is this TestObj used for?

core/src/lib.rs
```diff
@@ -1,15 +1,22 @@
+#![allow(unused_variables)]
+#![allow(dead_code)]
 mod credential;
-
 use credential::*;
 use serde_json::{self, Value};
+use signature::keypair::Ed25519SSIKeyPair;
 use std::collections::HashMap;
+// use registry_resolver::RegistryResolver;
```
LuisOsta, line 8: we should delete deadcode

core/src/lib.rs
```diff
@@ -1,15 +1,22 @@
...
 
 pub mod error;
 pub mod proof;
+use mockall::*;
+pub mod identity;
```
LuisOsta, line 13: can we move all `mod` definitions above the namespacing `use ...`

core/src/proof.rs
```diff
@@ -18,7 +18,10 @@ pub struct DataIntegrityProof {
 
 /// Given a JSON-LD document, create a data integrity proof for the document.
 /// Currently, only `Ed25519Signature2018` data integrity proofs in the JSON-LD format can be created.
-pub fn create_data_integrity_proof<S: signature::suite::Signature>(
+pub fn create_data_integrity_proof<
+    S: signature::suite::Signature,
+    // T: signature::keypair::Ed25519SSIKeyPair,
```
LuisOsta, line 23: Why is this commented out?

registry-resolver/Cargo.toml
```diff
@@ -1,9 +1,11 @@
 [package]
-name = "registry-resolver"
+name = "registry_resolver"
 version = "0.1.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+name = "registry_resolver"
```
LuisOsta, line 8: Is this necessary after renaming the package?

signature/src/signer.rs
```diff
@@ -1,5 +1,6 @@
 use crate::error::SignatureError;
 use crate::suite::{Ed25519Signature, Signature, VerificationRelation};
+// use std::convert::From;
```
LuisOsta, line 3: we should delete deadcode

signature/src/signer.rs
```diff
@@ -26,7 +27,7 @@ where
 }
 
 pub struct Ed25519DidSigner {
-    private_key: ed25519_zebra::SigningKey,
+    pub(crate) private_key: ed25519_zebra::SigningKey,
```
LuisOsta, line 30: The private key in the signer should not be public

core/src/identity.rs
```diff
@@ -0,0 +1,212 @@
+use serde::{Deserialize, Serialize};
+use signature::keypair::SSIKeyMaterial;
+
+use signature::keypair::Ed25519SSIKeyPair;
```
LuisOsta, line 4: We should not be using the specific implementation but instead the trait

core/src/identity.rs
```diff
@@ -0,0 +1,212 @@
...
+use signature::signer::DIDSigner;
+
+#[derive(Clone, Debug, Copy)]
+pub struct Identity<T> {
```
LuisOsta, line 8: Why make this a generic? DIDResolver is already a trait so I don't see the value here. We can just do:

```
pub struct Identity {
resolver: Box<dyn crate::DIDResolver>,
}
```

LuisOsta (changes requested):

core/src/identity.rs
mkatychev: Assuming functionality is the same, this would prevent a heap allocation by using `T` over `Box<T>`

sKudryashov: this change returns "the trait `DIDResolver` cannot be made into an object`DIDResolver` cannot be made into an object" error

core/src/proof.rs
sKudryashov: It's a leftover from the previous implementation

core/src/lib.rs
sKudryashov: It's a mock object and is used for applying a trait to it. In real case it has to be library-consumer owned object

core/src/identity.rs
sKudryashov: It's trait, but Since the KeyPair trait require public and private traits, we still have to maintain Ed25519SSIKeyPair or any other *KeyPair trait. There is still separation from Ed25519Signer traits.

core/src/identity.rs
```diff
@@ -0,0 +1,226 @@
...
+                    if did_doc_input.context != did_doc_mock.context {
+                        return false;
+                    }
+                    if did_doc_input.assertion_method[0].public_key_multibase
+                        != did_doc_mock.assertion_method[0].public_key_multibase
+                    {
+                        return false;
+                    }
+                    return true;
+                }),
```
mkatychev, line 200: these can be simplified into one statement: `return did_doc_input.id == did_doc_mock.id && did_doc_input.context == did_doc_mock.context && ...;`

core/src/identity.rs
```diff
@@ -0,0 +1,226 @@
...
+                }
+            }
+            Err(_) => {
+                if expect_ok {
+                    Err("Expected success".to_string())
+                } else {
+                    Ok(())
+                }
+            }
+        }
```
mkatychev, line 224: You can possible reduce nesting by using one match arm and an assertion:
https://github.com/mkatychev/darkroom/blob/9eaaec3bc63583125b67e2b1d25530385be790f9/src/record.rs#L307-L320

[`Result::is_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.is_err) is also good to panic on here depending on your match arm.

core/Cargo.toml
LuisOsta, line 21: We should delete this line

core/src/lib.rs
```diff
@@ -1,15 +1,18 @@
 mod credential;
+pub mod error;
+pub mod identity;
+pub mod proof;
 
 use credential::*;
+use mockall::*;
```
LuisOsta, line 7: delete this

core/src/lib.rs
```diff
@@ -1,15 +1,18 @@
...
 use std::collections::HashMap;
 
-pub mod error;
-pub mod proof;
-
 /// Verification of Data Integrity Proofs requires the resolution of the `verificationMethod` specified in the proof.
 /// The `verificationMethod` refers to a cryptographic key stored in some external source.
 /// The DIDResolver is responsible for resolving the `verificationMethod` to a key that can be used to verify the proof.
+
+#[automock]
```
LuisOsta, line 15: #[mockall::automock]

signature/src/keypair.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub(crate) master_private_key: ed25519_zebra::SigningKey,
+
+    pub(crate) authetication_public_key: ed25519_zebra::VerificationKey,
+    pub(crate) authetication_private_key: ed25519_zebra::SigningKey,
+
+    pub(crate) capability_invocation_public_key: ed25519_zebra::VerificationKey,
+    pub(crate) capability_invocation_private_key: ed25519_zebra::SigningKey,
+
+    pub(crate) capability_delegation_public_key: ed25519_zebra::VerificationKey,
+    pub(crate) capability_delegation_private_key: ed25519_zebra::SigningKey,
```
LuisOsta, line 69: We should investigate these linting issues

LuisOsta (changes requested):
A few minor issues but once they're done I'm happy to approve

LuisOsta (approved):

PR #43, opened 2022-08-10
19 implement restore identity

sKudryashov:


core/src/identity.rs
```diff
@@ -33,6 +33,16 @@ impl<T: crate::DIDResolver> Identity<T> {
         }
     }
 
+    pub async fn recover(
+        self,
+        did: String,
+        key_pair: Ed25519SSIKeyPair,
```
LuisOsta, line 39: Why not use the KeyPair trait instead of the specific one?

sKudryashov: KeyPair requires ed25519_zebra::SigningKey and ed25519_zebra::VerificationKey (or any other Private Public key set)

LuisOsta: Why does it require that? I thought we managed to design the KeyPair trait itself to be generic to such things?

sKudryashov: Yes but we can't do KeyPair without any specific algorithm. Our goal was to separate KeyPair from Signer, but generic KeyPair struct still has to accept the private and public key. Here is the conversion from KeyPair to signer: https://github.com/knox-networks/ssi/pull/40/files#diff-70d21711db4b5729408424f8caf1d0b5fce25c6abc89a2cc658585727e0a643eR5 and here is generic  KeyPair itself, https://github.com/knox-networks/ssi/pull/40/files#diff-1eb80c5452df3ad684ce36b4bfb4590a600978c10f0135bcba713ca18617d9bcR9 , we were implementing this on a pair code session and concluded there was no other way to have it implemented, pls correct me if I'm wrong.

core/src/identity.rs
```diff
@@ -1,15 +1,28 @@
 use serde::{Deserialize, Serialize};
 
-#[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
-pub struct KeyMaterial {
-    pub id: String,
-    #[serde(rename = "type")]
-    pub proof_type: String,
-    pub controller: String,
-    pub public_key_multibase: String,
+#[derive(Clone, Debug, Copy)]
```
LuisOsta, line 3: Can you delete this and have `recover` accept the `resolver` via parameters similar to `generate`? I believe we discussed the reasons why we should prefer having these functions be "pure" and lifecycle-less as part of the last PR

core/src/identity.rs
```diff
@@ -1,15 +1,28 @@
...
+    where
+        S: signature::suite::Signature,
+    {
+        let rsp = self.resolver.read(verifier.get_did()).await?;
+        Ok(rsp)
+    }
 }
 
-pub async fn generate<S: signature::suite::Signature>(
+pub async fn generate<S>(
```
LuisOsta, line 25: Do we wanna rename this to `create_identity`?

signature/src/keypair.rs
```diff
@@ -1,3 +1,6 @@
+#[warn(unused_imports)]
+use serde::{Deserialize, Serialize};
```
LuisOsta, line 2: Why don't we delete these if they are unused?

LuisOsta (changes requested):

core/src/identity.rs
```diff
@@ -1,28 +1,27 @@
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Copy)]
-pub struct Identity<T> {
-    resolver: T,
-}
+pub struct Identity {}
```
LuisOsta, line 3: We should delete this struct altogether as it's necessary. The user can just call the `recover` function directly

LuisOsta (changes requested):

core/src/identity.rs
```diff
@@ -1,15 +1,20 @@
...
-pub struct KeyMaterial {
-    pub id: String,
-    #[serde(rename = "type")]
-    pub proof_type: String,
-    pub controller: String,
-    pub public_key_multibase: String,
+pub struct Identity {}
+
+pub async fn recover<S, T>(
+    resolver: T,
```
LuisOsta, line 6: Is there a reason you made this a generic vs `impl crate::DidResolver` like with the verifier? Genuinely wanted to know if there was a reason for one vs the other in this case

LuisOsta (changes requested):

core/src/identity.rs
LuisOsta: @sKudryashov The change to make recover stateless looks good but we should still delete the `Identity` struct as it's not needed at least from what I see

core/src/identity.rs
LuisOsta: So apparently they are roughly the same from an implementation perspective - https://stackoverflow.com/questions/47514930/what-are-the-differences-between-an-impl-trait-argument-and-generic-function-par

LuisOsta: Resolving, though I'd still love to know if you had a preference for one or the other

PR #50, opened 2022-09-27
Integrate `generate_mnemonic` into `Ed25519SSIKeyPair`'s `new` function

LuisOsta:


LuisOsta:
I'm going to refactor some things to improve the error handling

signature/src/suite/ed25519_2020/ed25519_2020_error.rs
```diff
@@ -0,0 +1,11 @@
+#[derive(Debug, thiserror::Error)]
```
sKudryashov, line 1: Might be better to have  signature/src/suite/ed25519_2020/error.rs instead of signature/src/suite/ed25519_2020/ed25519_2020_error.rs because it is a part of ed25519_2020 mod already

sKudryashov (changes requested):

signature/src/suite/ed25519_2020/ed25519_2020_error.rs
LuisOsta: Okay sounds good, I had renamed it to something else to make searching easier but I'm having to change it to `error.rs`

sKudryashov (approved):

PR #61, opened 2022-10-07
DIDResolver changes necessary for CW

sKudryashov:
exposes some fields of did structs to the public  for 623-create-and-persist-identity-in-cw

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -13,24 +13,24 @@ use std::collections::HashMap;
 
 #[mockall::automock]
 #[async_trait::async_trait]
-pub trait DIDResolver {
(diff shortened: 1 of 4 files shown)


core/src/lib.rs
```diff
@@ -5,27 +5,31 @@ pub mod proof;
 
 use credential::*;
 use serde_json::{self, Value};
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::Debug};
```
LuisOsta, line 8: We should remove debug here

LuisOsta (changes requested):
One minor issue but overall looks good

LuisOsta (approved):
LGTM

PR #65, opened 2022-10-13
Create & Persist Identity As Part Of Wallet Creation In Custodial Wallet#623

sKudryashov:
SSI part of https://github.com/knox-networks/core/issues/623


registry_resolver/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-mod registry_client;
+pub mod registry_client;
```
LuisOsta, line 1: This shouldn't be public

LuisOsta (changes requested):

registry_resolver/src/lib.rs
sKudryashov: this is part of trait declaration on server.rs here, not sure it can be avoided https://github.com/knox-networks/core/pull/693/files#diff-4c7161a5afc4e79f458609693cf8b7b4dd07db81b67884e7c444ba626253b1feR453

LuisOsta (approved):

PR #78, opened 2022-10-20
Create did doc test added

sKudryashov:


core/src/identity.rs
```diff
@@ -176,6 +180,50 @@ mod tests {
         String::from("123456789")
     }
 
+    #[ignore]
```
LuisOsta, line 183: You should remove this `#[ignore]`

LuisOsta (changes requested):

core/src/identity.rs
```diff
@@ -176,6 +180,36 @@ mod tests {
         String::from("123456789")
     }
 
+    #[rstest::rstest]
+    #[case::creates_successfully()]
+    fn test_create_did_doc() -> Result<(), String> {
+        let kp = signature::suite::ed25519_2020::Ed25519KeyPair::new(None).unwrap();
```
LuisOsta, line 186: So if we are going to do a structural comparison vs a complete comparison we should make it clear that the expectation json is a template and not an actual DID Document itself

LuisOsta (changes requested):

core/src/identity.rs
LuisOsta: Alternatively we can do a complete comparison by making the mnemonic fixed and restoring the KeyPair from that (instead of creating a new one

mkatychev (approved):

LuisOsta (approved):

PR #90, opened 2022-10-28
88 cannot deserialize verifiable credential

sKudryashov:


LuisOsta (commented):
I think we should have at least one basic test to verify that deserialization works

PR #110, opened 2022-12-16
109 ffi ssi create and push the did doc to registry mgmt

sKudryashov:
This PR enables FFI in SSI and exposes two SSI actions: `create did document` and `register did document` via the corresponding SSI API. In `register did document` there is an external call to registry-management using GRPC. On Rust side there is an FFI shim, on C side there are two unity tests to ensure we can call `create did document` and `register did document` from C. 
We build .a and .h files along with dylib here as well, as described in https://github.com/knox-networks/core/issues/1214


fkim7:
@sKudryashov Can we have the SSI has a separate dylib so we dont bloat the current monetae dylib? @mkatychev

ffi/src/did.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[ffi_export]
+pub fn create_identity(
+    rust_error: MaybeRustError,
+    did_method: char_p::Ref<'_>,
+    mnemonic_input: char_p::Ref<'_>,
+) -> Option<repr_c::Box<DidDocument>> {
+    // create_did_doc
+    let _span = error_span!("ffi_ssi").entered();
+    info!(
```
fkim7, line 22: @sKudryashov how do we get these logs to carry over to the C side and eventually Dart?

sKudryashov: My assumption is that it should work out the same way as `println!()` macro, since both are macros. println! works and prints an output, info! most likely will require some pre-configuration on the shim side.

Cargo.toml
```diff
@@ -5,4 +5,39 @@ members = [
...
 	"ssi",
 	"ephemeral_resolver",
+	"ffi"
 ]
+
+[workspace.package]
+version = "0.3.0"
+edition = "2021"
+
+[workspace.dependencies]
```
LuisOsta, line 15: Are all of these workspace deps needed?

registry_resolver/Cargo.toml
```diff
@@ -8,14 +8,18 @@ edition = "2021"
 [dependencies]
 prost = "0.11.0"
 prost-types = "0.11.0"
-tonic = "0.8.0"
+tonic = { version = "0.8.2", features = ["tls", "tls-roots", "tls-webpki-roots"] }
```
LuisOsta, line 11: Why are these new imports needed?

sKudryashov: They are needed to enable grpc over https

fkim7:
@sKudryashov is it possible to generate .a files for libmonetae_ffi as well?  @yanctrindade needs to replace all dylibs with .a files for his current experiment.

sKudryashov:
> @sKudryashov is it possible to generate .a files for libmonetae_ffi as well? @yanctrindade needs to replace all dylibs with .a files for his current experiment.

Yes, of course

fkim7 (approved):

LuisOsta (approved):
I don't know if we need all of those workspace deps but other than that LGTM

PR #111, opened 2023-01-05
105 Add Context-Based Validation To VC & VP Creation

LuisOsta:


core/src/lib.rs
```diff
@@ -331,4 +343,107 @@ mod tests {
...
+                        let name = object
+                            .as_node()
+                            .unwrap()
+                            .get_any(&iri!("http://xmlns.com/foaf/0.1/name"))
+                            .unwrap()
+                            .as_str()
+                            .unwrap();
+
+                        println!("id: {id}");
+                        println!("name: {name}");
```
sKudryashov, line 441: Looks like debugging output as well

core/src/lib.rs
```diff
@@ -331,4 +343,107 @@ mod tests {
...
+                    }
+                }
+            }
+        }
+
+        // assert!(false);
```
sKudryashov, line 447: What assertions do we expect in this test, i.e. what are we testing in particular? Could you pls add?

sKudryashov (changes requested):

core/src/lib.rs
LuisOsta: This test is being ignored, its meant to be followed back on once we actually are able to finish this issue completely. This PR is to include all of the other fixes made while putting work on this issue

sKudryashov (approved):

Issue #117, opened 2023-02-14
DID Doc Generation on SSI Library


yanctrindade:
Did doc generated from SSI Library has some difference/issue against the old one generated from monetae lib and that's causing a bug when trying to associate a new wallet.

that os exposed on https://github.com/knox-networks/dart-sdk/issues/53

sKudryashov:
https://github.com/knox-networks/dart-sdk/issues/53

PR #121, opened 2023-02-28
@context record added

sKudryashov:
Fixes https://github.com/knox-networks/ssi/issues/117, the only difference between monetae and ssi is the @context field.

### DIFF
--- a/core/src/identity.rs
+++ b/core/src/identity.rs
@@ -40,7 +40,10 @@ where
     S: signature::suite::Signature,
 {
     DidDocument {
-        context: vec!["https://www.w3.org/ns/did/v1".to_string()],
+        context: vec![
+            "https://www.w3.org/ns/did/v1".to_string(),
+            "https://w3id.org/security/suites/ed25519-2020/v1".to_string(),
+        ],
         id: verifier.get_did(),
         authentication: vec![KeyMaterial {
             id: verifier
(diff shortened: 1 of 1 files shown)


LuisOsta (approved):

fkim7 (approved):

yanctrindade (changes requested):

core/src/identity.rs
```diff
@@ -40,8 +40,11 @@ where
     S: signature::suite::Signature,
 {
     DidDocument {
-        context: vec!["https://www.w3.org/ns/did/v1".to_string()],
         id: verifier.get_did(),
+        context: vec![
+            "https://www.w3.org/ns/did/v1".to_string(),
+            "https://w3id.org/security/suites/ed25519-2020/v1".to_string(),
+        ],
```
fkim7, line 47: @sKudryashov @LuisOsta  Can we move these URLs out of the code and at least into a config file? we can have a separate ticket for this outside of this PR but if the URL changes we'd have to update the builds/images and deliver new helm charts to all customers instead of our professional services team/customer team be able to make some updates by configuration.  Or maybe since SSI is shipped as an open source library its less of a concern compared to applications?

sKudryashov: I think in case of using it as an SSI library entirely probably we don't need an initialization from a config or something, rather we have to accept it as a mandatory parameter, @LuisOsta what do you think?

sKudryashov: Thus we can put this maintain burden on the library consumer rather then us which is the right thing to do for a library

LuisOsta: Let's revisit this as part of a different issue as I want to make sure there aren't usage or operational implications that would heavily affect this consideration

fkim7: created https://github.com/knox-networks/ssi/issues/122

yanctrindade (approved):

fkim7 (approved):

PR #144, opened 2023-08-21
143 Update Registry Resolver Client With Spec Compliant Version

LuisOsta:


LuisOsta:
@adewaleafolabi In the client right now we just return the Did Document, what do you think the value is of returning the other things that are part of the `resolve` response?

LuisOsta:
The steps after this PR is done is to:

- Create a new SSI release
- Update CW to work with new release (setup PR)
- Merge Gomono Registry Mgmt PR
- Merge CW PR

adewaleafolabi:
> @adewaleafolabi In the client right now we just return the Did Document, what do you think the value is of returning the other things that are part of the `resolve` response?

The resolve response according to the spec must contain three objects 

- didResolutionMetadata
- didDocument
- didDocumentMetadata

image

For SSI, if we don't need those other items we can just skip it for now. They are basically metadata

LuisOsta:
@adewaleafolabi Alright sounds good, the question was more about what we did or didn't want to expose. For something this low level it makes sense to expose everything

LuisOsta:
@adewaleafolabi Are the values in `didUrl` in resolution metadata optional or required?

adewaleafolabi (approved):

PR #145, opened 2023-08-24
use protofetch instead of buf

gshuflin:
Swap over the ssi repo from buf to protofetch. This PR also has the effect of bumping the version number of the entire SSI workspace to `0.4.0`, and consistently using the workspace version for the versions of the various subcomponent packages of this project.


.github/workflows/build.yaml
```diff
@@ -31,6 +29,10 @@ jobs:
         run: |-
           cargo install cargo-make
 
+      - name: Installing protofetch
+        run: |-
+          cargo install --git https://github.com/knox-networks/protofetch --branch tracing
```
mkatychev, line 34: Reposting:
we should stash a binary in the releases tag
so we can `wget` instead of spending a while building from source

.github/workflows/build.yaml
```diff
@@ -31,6 +32,16 @@ jobs:
...
 
+      - name: Installing protofetch
+        # run: |-
+        #   cargo install protofetch
+        run: |-
+          cargo install --git https://github.com/knox-networks/protofetch --branch nonzero-error-code
+
+      - name: Setup git dependency
+        run: |-
+          git config credential.helper '!f() { echo username=$EXTERNAL_GIT_USER; echo password=$EXTERNAL_GIT_TOKEN; };f'
```
kbeckman, line 43: @gshuflin -- I forgot that we already have a `developers@knox-networks.com` GitHub account which exists for this kind of thing... I created a PAT for this account that should give repo access to any repo in the `knox-networks` organization as well as public repos. I saved the PAT as an organizational GHA secret: `PROTOFETCH_GITHUB_TOKEN`. You should be able to hit it within a GHA workflow context by using `{{ secrets.PROTOFETCH_GITHUB_TOKEN }}`.

Once you work out the specifics here, I'd like to see the protofetch code moved into and documented as a reusable action across workflows in `knox-networks/github-actions`. That way you won't have to copy/past so much code between projects. Hit me up if you need help getting started with that...

jkleinknox:
So I see you merged some changes into the upstream repo. I assume we use the upstream version and not the forked version of protofetch?

gshuflin:
> So I see you merged some changes into the upstream repo. I assume we use the upstream version and not the forked version of protofetch?

Yeah, we're no longer using our own fork. The github action is pulling a specific git hash from the upstream repo: https://github.com/knox-networks/github-actions-public/blob/main/.github/actions/setup-protofetch/action.yml#L19 that has the changes we want.

registry_resolver/build.rs
```diff
@@ -1,14 +1,63 @@
...
+        "protofetch stderr: {}",
+        String::from_utf8_lossy(&stderr).into_owned()
+    );
+
     if !status.success() {
-        return Err(String::from_utf8_lossy(&stderr).into());
+        return Err(BuildError::with_kv(
+            "error_msg",
+            String::from_utf8_lossy(&stderr).into_owned(),
+        ));
```
mkatychev, line 35:
```suggestion
        return Err(BuildError::attach(String::from_utf8_lossy(&stderr).into_owned()));
```
Not sure we need a kv here

registry_resolver/src/registry_client.rs
```diff
@@ -1,5 +1,5 @@
 pub mod registry {
-    include!(concat!(env!("OUT_DIR"), "/registry_api.v1.rs"));
+    include!(concat!(env!("OUT_DIR"), "/gen/registry_api.v1.rs"));
```
mkatychev, line 2: Have we verified that `registry_api.v1.rs` is the right file structure? I'm pretty sure `tonic-build` doesn't produce a `.v1` suffix (infact rust takes issue with full points in rust files other than the one in `.rs`).

gshuflin: After doing a clean build:
```
➤ exa -T target/debug/build/registry_resolver-*
target/debug/build/registry_resolver-29dfdb74e79f4851
├── invoked.timestamp
├── out
│  ├── gen
│  │  └── registry_api.v1.rs
│  ├── proto_descriptor.bin
│  └── protos
│     └── registry-mgmt
│        └── registry_api
│           └── v1
│              └── registry.proto
├── output
├── root-output
└── stderr
target/debug/build/registry_resolver-eac1c387d1002808
├── build-script-build
├── build_script_build-eac1c387d1002808
└── build_script_build-eac1c387d1002808.d
```

so it looks like the tonic code is in fact generating a rust source file with a period in it, which doesn't seem to be breaking anything as far as I can tell.

.github/workflows/build.yaml
```diff
@@ -31,6 +29,11 @@ jobs:
         run: |-
           cargo install cargo-make
 
+      - uses: knox-networks/github-actions-public/.github/actions/setup-protofetch@main
```
mkatychev, line 32: Should we use `wget` or `curl` to pull in a prebuilt binary?
If we plan to use this change downstream, building `protofetch` will add roughly two minutes to the CI time:
 https://github.com/knox-networks/ssi/actions/runs/6052152827/job/16425074210?pr=145

.github/workflows/build.yaml
```diff
@@ -31,6 +29,11 @@ jobs:
         run: |-
           cargo install cargo-make
```
mkatychev, line 30: Should we pull in the prebuilt binary here similar to what is done below?
https://github.com/knox-networks/ssi/blob/e2d20423de4ce5fa1f98c891cf6e604a69a964b9/.github/workflows/release_ffi.yaml#L75
```suggestion
        run: ./container/deps.sh --force cargo-make
```

registry_resolver/src/registry_client.rs
```diff
@@ -1,5 +1,5 @@
 pub mod registry {
-    include!(concat!(env!("OUT_DIR"), "/registry_api.v1.rs"));
+    include!(concat!(env!("OUT_DIR"), "/gen/pb/registry_api.v1.rs"));
```
mkatychev, line 2: Correct me if I'm wrong but [`OUT_DIR` implies generated code](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates).
```suggestion
    include!(concat!(env!("OUT_DIR"), "/pb/registry_api.v1.rs"));
```

registry_resolver/build.rs
```diff
@@ -1,14 +1,58 @@
...
+        return Err(BuildError::attach(
+            String::from_utf8_lossy(&stderr).into_owned(),
+        ));
     }
+
+    // The following protobuf binding setup is adapted from what is documented at
+    // https://docs.rs/pbjson-build/latest/pbjson_build/  - note that `tonic_build`
+    // accepts many of the same arguments as `prost_build` does in the documented example.
+
+    let generated_code_dir = PathBuf::from(&out_dir).join("gen").join("pb");
```
mkatychev, line 41: If https://github.com/knox-networks/ssi/pull/145/files#r1313366805 has merit:
```suggestion
    let generated_code_dir = PathBuf::from(&out_dir).join("pb");
```

mkatychev (approved):

mkatychev (approved):
LGTM

registry_resolver/src/registry_client.rs
gshuflin: The documentation says:
```
OUT_DIR — the folder in which all output and intermediate artifacts should be placed. This folder is inside the build directory for the package being built, and it is unique for the package in question.
```
which implies to me that there is a distinction between "output" and "intermediate" artifacts. So I think there's value in having a separate `gen` directory that is specifically for output artifacts that need to get imported into code. In fact `protofetch` puts its protobuf directory structure, which we can think of as an intermediate artifact, into `OUT_DIR` but not under `gen`, so someone looking at this code later can assume that `gen` only has rust sourcefiles and the other stuff in the `OUT_DIR` directory is not immediately relevant to them if they just care about the generated rust. So I think keeping the `gen` makes sense.

mkatychev: makes perfect sense

PR #153, opened 2023-08-30
Update README.md FFI build steps

fkim7:


fkim7:
@LuisOsta @yanctrindade actually we'll need something here about the process to push the ffi artifacts upstreams

PR #155, opened 2023-09-05
Changes to facilitate building within nix

gshuflin:
This PR does two things, to facilitate `ssi` being built within a nix flake:
* use a local cache directory within $OUT_DIR for the protofetch cache, instead of a global cache
* expose a `PROTOFETCH_OVERRIDE_REPOSITORY` env var, which will instruct the internal protofetch process to fetch the protobuf from a local file instead of trying to directly fetch it from github

### DIFF
--- a/registry_resolver/build.rs
+++ b/registry_resolver/build.rs
@@ -9,20 +9,30 @@ fn main() -> Result<(), Report<BuildError>> {
     let protofetch_cache_dir = PathBuf::from(&out_dir).join("protofetch_cache_dir");
     std::fs::create_dir_all(&protofetch_cache_dir).report_as()?;
 
+    let mut command = std::process::Command::new("protofetch");
+
+    command
+        .env("RUST_LOG", "TRACE")
+        .arg("--output-proto-directory")
+        .arg(&proto_dir)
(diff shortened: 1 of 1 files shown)


registry_resolver/build.rs
```diff
@@ -3,21 +3,36 @@ use std::{env, path::PathBuf};
...
+        .env("RUST_LOG", "TRACE")
+        .arg("--output-proto-directory")
+        .arg(&proto_dir)
+        .arg("--cache-directory")
+        .arg(&protofetch_cache_dir);
+
+    command.arg("fetch");
+
+    if let Ok(dir) = env::var("PROTOFETCH_OVERRIDE_REPOSITORY") {
+        println!("Value of PROTOFETCH_OVERRIDE_REPOSITORY: {dir}");
```
mkatychev, line 24: should we use the `cargo:warning=` prefix here?
```suggestion
        println!("cargo:warning=PROTOFETCH_OVERRIDE_REPOSITORY=\"{dir}\"");
```

mkatychev (approved):

LuisOsta (approved):

PR #158, opened 2023-09-22
Clean up dependencies

mkatychev:
Handle dependabot warnings: https://github.com/knox-networks/ssi/security/dependabot/7

* Removed unused dependencies with `cargo-udeps`:
  ```
  $ cargo +nightly udeps --all-targets

  unused dependencies:
  `ssi-ffi v0.4.0 (/Users/mkatychev/Documents/knox/ssi/ffi)`
  └─── dependencies
       └─── "ephemeral_resolver"
  `ssi_core v0.4.0 (/Users/mkatychev/Documents/knox/ssi/core)`
  └─── dev-dependencies
       ├─── "async-std"
       └─── "map"
  Note: They might be false-positive.
        For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
        To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.

  ```
  
* Run [`cargo upgrade`](https://github.com/killercup/cargo-edit#cargo-upgrade)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,23 +15,23 @@ members = [
 
 
 [workspace.dependencies]
-base64 = "0.13.0"
-chrono = { version = "0.4.19", features = ["serde"] }
-futures = { version = "0.3.21", default-features = false }
-mockall = { version = "0.11.2" }
-multibase = "0.9.0"
-prost = "0.11.0"
-prost-types = "0.11.0"
-serde = { version = "1.0.136", features = ["derive"] }
-serde_json = "1.0.85"
+base64 = "0.13.1"
+chrono = { version = "0.4.31", features = ["serde"] }
+futures = { version = "0.3.28", default-features = false }
+mockall = { version = "0.11.4" }
+multibase = "0.9.1"
+prost = "0.11.9"
+prost-types = "0.11.9"
+serde = { version = "1.0.188", features = ["derive"] }
+serde_json = "1.0.107"
 signature = "1.6.4"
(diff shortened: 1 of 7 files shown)


LuisOsta (approved):
@mkatychev Overall looks good to me but I'd prefer we upgrade the `signature` crate separately so we can do a deep review of the chances to make sure there aren't any semantics or nuances that have changed that we are relying on. I call out this crate specifically cause cryptography is an area that I think is ideal to be more sensitive around and its best to play it safer with that

Cargo.toml
```diff
@@ -15,23 +15,23 @@ members = [
...
+base64 = "0.13.1"
+chrono = { version = "0.4.31", features = ["serde"] }
+futures = { version = "0.3.28", default-features = false }
+mockall = { version = "0.11.4" }
+multibase = "0.9.1"
+prost = "0.11.9"
+prost-types = "0.11.9"
+serde = { version = "1.0.188", features = ["derive"] }
+serde_json = "1.0.107"
 signature = "1.6.4"
```
mkatychev, line 27: @LuisOsta  signature crate has not been updated

mkatychev:
@LuisOsta  are you talking about  https://docs.rs/signature/latest/signature/ or the workspace crate?

LuisOsta:
@mkatychev I'm talking about the workspace crate `signature`

LuisOsta:
Specifically `ed25519-zebra` and `sha2` were updated

PR #162, opened 2023-11-16
29 Add Support RDFC-1.0 Normalization In Data Integrity Proofs

LuisOsta:
-  Adds RDFC-1.0 normalization support in place of URDNA2015 (as they are essentially equivalent for our purposes - https://www.w3.org/TR/rdf-canon/#urdna2015) to add complete support for legacy `Ed25519SignatureSystem` support https://www.w3.org/TR/vc-di-eddsa/#the-ed25519signature2020-suite. In future work we'll need to support the new `eddsa-rdfc-2022` and `DataIntegrityProof` approach

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -454,6 +454,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a26ae43d7bcc3b814de94796a5e736d4029efb0ee900c12e2d54c993ad1a1e07"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


fkim7:
> * Adds RDFC-1.0 normalization support in place of URDNA2015 (as they are essentially equivalent for our purposes - https://www.w3.org/TR/rdf-canon/#urdna2015) to add complete support for legacy `Ed25519SignatureSystem` support https://www.w3.org/TR/vc-di-eddsa/#the-ed25519signature2020-suite. In future work we'll need to support the new `eddsa-rdfc-2022` and `DataIntegrityProof` approach

Just a note to self, RDFC-1.0 should be compatible with URDNA2015 as well

fkim7 (approved):

PR #167, opened 2023-11-22
166 integrate vc data model spec test suite into ssi

LuisOsta:


cli/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+    },
+}
+
+fn main() {
+    let args = CliArguments::parse();
+    match args.command {
+        Command::TestIssueCredential { input_file_path } => {
+            //read the JSON string file in input_file_path and parse it into a VerifiableCredential
+            //print the VerifiableCredential
+            let input = std::fs::read_to_string(input_file_path).unwrap();
```
gshuflin, line 31: Is it worthwhile to handle errors like this in a nicer way than just panicking (e.g. have the entire `main()` function return some kind of formatted error type)? Who do we see as the ultimate audience for this cli tool?

cli/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+        input_file_path: String,
+    },
+}
+
+fn main() {
+    let args = CliArguments::parse();
+    match args.command {
+        Command::TestIssueCredential { input_file_path } => {
+            //read the JSON string file in input_file_path and parse it into a VerifiableCredential
+            //print the VerifiableCredential
```
gshuflin, line 30: `clap` will let you autogenerate command line help text from doc comments (see the example on https://docs.rs/clap/latest/clap/) - maybe this comment could live on the definition of `TestIssueCredential` as a doc comment.

LuisOsta: That's great to know! Let me see

cli/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+struct CliArguments {
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Subcommand, Debug)]
+enum Command {
+    TestIssueCredential {
+        #[arg(short, long)]
+        input_file_path: String,
```
gshuflin, line 17: since these names ultimately become command-line arguments/flags, it'd be good to make them shorter I think. this could just be `input_file`

core/src/credential.rs
```diff
@@ -20,6 +32,78 @@ pub enum CredentialType {
...
+    Set(Vec<std::collections::HashMap<String, serde_json::Value>>),
+}
+
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone, Validate)]
+pub struct Credential {
+    #[validate(custom(validation::credential_context_validation))]
+    #[serde(rename = "@context")]
+    pub context: DocumentContext,
+
+    #[serde(rename = "id")]
```
gshuflin, line 59: Is this rename directive necessary? The field itself is already called `id`

cli/src/main.rs
LuisOsta: Done

core/src/credential.rs
```diff
@@ -20,6 +32,78 @@ pub enum CredentialType {
...
     BankCard,
     BankAccount,
+    UniversityDegreeCredential,
+    AlumniCredential,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, Eq)]
+pub enum PresentationType {
+    #[serde(rename = "VerifiablePresentation")]
+    Common,
```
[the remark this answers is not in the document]
LuisOsta: I can do so if I add a clarifying comment

core/src/credential.rs
LuisOsta: This is a holdover from the @id days, I think its something we can remove

cli/src/main.rs
LuisOsta: @gshuflin For the foreseeable future this would just be for internal use, I can imagine a future where this could be used in other contexts theoretically but not for a while if at all.

What would be the benefit of handling this more nicely? This CLI would always exit after a command is finished

core/src/credential.rs
```diff
@@ -38,41 +124,36 @@ impl CredentialType {
             "BankAccount" => Some(CredentialType::BankAccount),
             "PermanentResidentCard" => Some(CredentialType::PermanentResidentCard),
             "VerifiableCredential" => Some(CredentialType::Common),
+            "UniversityDegreeCredential" => Some(CredentialType::UniversityDegreeCredential),
+            "AlumniCredential" => Some(CredentialType::AlumniCredential),
             _ => None,
```
gshuflin, line 133: I like to write this sort of pattern as:
```
Some(match cred_type => {
   "BankCard => CredentialType::BankCard
   ...
   _ => return None
   })
```

since it reduces the number of times you have to write out `Some`

core/src/credential.rs
```diff
@@ -105,19 +186,29 @@ impl Credential {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let vp = VerifiablePresentation::from_json_str(s)?;
+        Ok(vp)
+    }
+}
+
+impl std::fmt::Display for VerifiablePresentation {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let vp = self.to_json_string().unwrap();
```
gshuflin, line 213: same thing here, should be be crashing if this unwrap fails?

core/src/credential.rs
LuisOsta: @gshuflin That makes sense, also do you konw if there's a way to guarantee that there's no members of `CredentialType` left unmatched?

core/src/credential/validation.rs
```diff
@@ -0,0 +1,32 @@
...
+) -> Result<(), serde_valid::validation::Error> {
+    if val.is_empty() {
+        return Err(serde_valid::validation::Error::Custom(
+            "Context must contain at least one URI".to_string(),
+        ));
+    }
+
+    match val[0] {
+        ContextValue::String(ref s) => {
+            if s != super::BASE_CREDENTIAL_CONTEXT {
```
gshuflin, line 16: This check can be part of the match statement itself:

```
ContextValue::String(ref s) if s != super::BASE_CREDENTIAL_CONTEXT => {
   return Err(...)
```

core/src/credential.rs
LuisOsta: My one issue with this approach is that it makes the return model more complicated/unclear

core/src/credential/validation.rs
```diff
@@ -0,0 +1,32 @@
...
+pub fn credential_context_validation(
+    val: &Vec<ContextValue>,
+) -> Result<(), serde_valid::validation::Error> {
+    if val.is_empty() {
+        return Err(serde_valid::validation::Error::Custom(
+            "Context must contain at least one URI".to_string(),
+        ));
+    }
+
+    match val[0] {
```
gshuflin, line 14: I'd probably craft this whole thing as 
```
match val[0] {
   error_condition => Err(...),
   error_condition2 => Err(...),
   _ => Ok(())
}
```

to avoid the explicit `return`s. You could make this whole method even  more concise by `match`-ing on `val.get(0)` and treating a `None` return value as the first "context must contain at least one URI" error intead of making that be a separate check

core/src/credential/validation.rs
LuisOsta: @gshuflin The problem with this is that then I have to create another match arm to cover where the context value is a string but not the conditional isn't, which ends up being longer than if I just had this

core/src/credential/validation.rs
LuisOsta: Okay let me see about this

core/src/proof.rs
```diff
@@ -16,11 +17,40 @@ pub struct DataIntegrityProof {
     pub proof_value: String,
 }
 
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone, PartialEq, Eq)]
+pub struct RsaSignature2018 {
+    #[serde(rename = "type")]
+    pub proof_type: String,
+    #[serde(rename = "created")]
```
gshuflin, line 24: Is this rename necessary?

core/src/proof.rs
```diff
@@ -16,11 +17,40 @@ pub struct DataIntegrityProof {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub created: Option<String>,
+    #[serde(rename = "verificationMethod")]
+    pub verification_method: String,
+    #[serde(rename = "proofPurpose")]
+    pub proof_purpose: String,
+    #[serde(rename = "jws")]
```
gshuflin, line 31: same here

core/src/proof.rs
```diff
@@ -16,11 +17,40 @@ pub struct DataIntegrityProof {
...
+    pub jws: String,
+}
+
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum CredentialProof {
```
gshuflin, line 37: Why not just have this be a type alias for `Vec<ProofType>` and allow the vec to contain one thing? Do we need a way to statically guarantee that there are never 0 credential proofs?

core/src/credential/validation.rs
LuisOsta: This solution actually does make a lot of sense

core/src/proof.rs
LuisOsta: I don't think so

core/src/proof.rs
LuisOsta: I don't think so

core/src/proof.rs
```diff
@@ -87,21 +119,29 @@ mod tests {
...
-                    signer.get_verification_method(relation)
-                );
-                assert_eq!(proof.proof_purpose, relation.to_string());
-                let transformed_data =
-                    crate::proof::normalization::create_hashed_normalized_doc(doc).unwrap();
-                let mut hasher = sophia::c14n::hash::Sha256::initialize();
-                hasher.update(&transformed_data);
-                let hash_data = hasher.finalize();
+                if let super::CredentialProof::Single(proof) = proof {
+                    if let super::ProofType::Ed25519Signature2020(proof) = proof {
```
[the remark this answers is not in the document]
LuisOsta: I knew that there had to be a better way, thanks!

core/src/proof.rs
LuisOsta: @gshuflin Credentials proofs can be structured as either a array of objects or just a singular object. And we need to be able to parse them and handle them in either scenario.

If we only expect an array we won't be able to parse the proofs where its just an object. This is part of the nature of the VC spec

core/src/credential.rs
```diff
@@ -38,41 +124,36 @@ impl CredentialType {
...
-pub type CredentialSubject = std::collections::HashMap<String, serde_json::Value>;
-
-#[derive(Debug, serde::Serialize, serde::Deserialize, Clone)]
-pub struct VerifiableCredential {
-    #[serde(flatten)]
-    pub credential: Credential,
-    pub proof: crate::proof::DataIntegrityProof,
+impl std::fmt::Display for VerifiableCredential {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let vc = self.to_json_string().unwrap();
```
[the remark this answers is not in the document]
LuisOsta: Let me think about this, this would only fail if we had somehow constructed a Verifiable Credential or VP that was incorrect on validation. Which should be impossible. So I lean towards keeping the panic here, but let me think about it

core/src/credential.rs
LuisOsta: See comment here - https://github.com/knox-networks/ssi/pull/167#discussion_r1406883667

cli/src/main.rs
gshuflin: Panicking if you can't do something like open a file passed into a command line app is kind of unfriendly (as opposed to a nicely-formatted error message), but if the only people using this tool for the forseeable future will be us then I probably wouldn't spend that much time trying to clean it up now.

LuisOsta: Yeah I agree its unfriendly but like you said as an internal tool I don't think its an urgent/valuable enough issue right now

core/src/credential.rs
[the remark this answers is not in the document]
LuisOsta: Done

cli/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#[derive(Subcommand, Debug)]
+enum Command {
+    /// Read the JSON string file in input_file_path and parse it into a VerifiableCredential. Then prints the VerifiableCredential
+    TestIssueCredential {
+        #[arg(short, long)]
+        input_file: String,
+    },
+    /// Read the JSON string file in input_file_path and parse it into a VerifiablePresentation. Then prints the VerifiablePresentation
```
gshuflin, line 20: Change the comment here (and the one above) to `input_file` instead of `input_file_path`

core/src/credential.rs
```diff
@@ -75,6 +76,102 @@ pub struct Credential {
...
+            CredentialType::UniversityDegreeCredential => "UniversityDegreeCredential",
+            CredentialType::AlumniCredential => "AlumniCredential",
+        }
+    }
+}
+
+impl std::fmt::Display for VerifiableCredential {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.to_json_string() {
+            Ok(vc) => {
```
gshuflin, line 147: Nit: you don't need the surrounding `{ }` this can just be `Ok(vc) => write!(f, "{}", vc)`, and same for the Err case

core/src/credential.rs
```diff
@@ -75,6 +76,102 @@ pub struct Credential {
...
+        }
+    }
+}
+
+impl FromStr for Credential {
+    type Err = super::error::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let c = Credential::from_json_str(s)?;
+        Ok(c)
```
gshuflin, line 162: This can just be a bare `Credential::from_json_str`, and if this repo had clippy configured the same way that `core` does I think it would emit a style warning for writing it this way

core/src/credential.rs
```diff
@@ -105,19 +202,35 @@ impl Credential {
...
-pub struct VerifiablePresentation {
-    #[serde(flatten)]
-    pub presentation: Presentation,
-    pub proof: crate::proof::DataIntegrityProof,
+impl FromStr for Presentation {
+    type Err = super::error::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let p = Presentation::from_json_str(s)?;
+        Ok(p)
```
gshuflin, line 210: same here, in general:

```
let a = some_func()?;
Ok(a)
```

can always be transformed into:
```
some_func()
```

(barring weirdness with `?` doing error type conversions)

core/src/credential.rs
```diff
@@ -105,19 +202,35 @@ impl Credential {
...
-    #[serde(rename = "@context")]
-    pub context: VerificationContext,
-    #[serde(rename = "verifiableCredential")]
-    pub verifiable_credential: Vec<VerifiableCredential>,
+impl FromStr for VerifiablePresentation {
+    type Err = super::error::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let vp = VerifiablePresentation::from_json_str(s)?;
+        Ok(vp)
```
gshuflin: same here

core/src/credential.rs
```diff
@@ -105,19 +202,35 @@ impl Credential {
...
+
+impl std::fmt::Display for VerifiablePresentation {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.to_json_string() {
+            Ok(vp) => {
+                write!(f, "{}", vp)
+            }
+            Err(e) => {
+                write!(f, "Error: {}", e)
+            }
```
gshuflin: braces unnecessary here too

gshuflin (approved):

core/src/credential.rs
fkim7: Unrelated to this PR but we need to refactor this to avoid needing to hardcode the credentialTypes.  We need to be able to support various new cred types- ex DriversLicense, ProofOfEmployment, etc without needing to issue new SSI SDKs and clients needing to do new builds... Note Registry Mgmt will have all VC schema types so we can refer to that as well if required for schema definitions.

PR #172, opened 2023-12-07
163 Add Integrate Test Vectors For Data Integrity Proofs

LuisOsta:
- Created GH Action To Run vc-di-eddsa data integrity test vectors for Ed25519Signature2020 - https://github.com/w3c/vc-di-eddsa/tree/main/TestVectors
- Created new command for CLI to run the Ed25519Signature2020 data integrity tests
- Refactored signature suite to support keypair creation from private key & serialization of VerificationRelation
- Fixed Ed25519 DataIntegrity algorithm in SSI Core. Supporting V2 Data Model for data integrity tests


gshuflin:
Can you add a more detailed description of what this PR is doing?

.github/workflows/data_integrity_spec_test.yaml
```diff
@@ -0,0 +1,44 @@
...
+          cross-repo-token: ${{ secrets.PROTOFETCH_GITHUB_TOKEN }}
+
+      - name: Test protos
+        run: (cd registry_resolver; protofetch fetch) && git diff --exit-code
+
+      - name: Build CLI
+        run: cargo build --package cli --release --features="v2_test"
+
+      - name: Add CLI to bin
+        run: sudo cp $PWD/target/release/ssi_cli /bin/ssi_cli
```
gshuflin, line 28: What if we skipped this step and just ran `ssi_cli` from this `$PWD/target/release/ssi_cli` path in the step below?

LuisOsta:
@gshuflin Let me know if what I've added provides the desired detail

.github/workflows/data_integrity_spec_test.yaml
LuisOsta: @gshuflin For this it might be possible but I'd prefer to follow the same model as the other spec test (which did require it to be part of the bin folder

core/src/proof.rs
```diff
@@ -53,6 +78,18 @@ impl std::fmt::Display for DataIntegrityProof {
     }
 }
 
+impl ProofOptionDocument {
```
gshuflin, line 81: Any reason not to just implement `From<ProofOptionDocument>` for `DataIntegrityProof`?

LuisOsta: @gshuflin This requires an input, its not a pure conversion so we can't do `From<ProofOptionDocument>` since that requires no parameters to be passed?

LuisOsta: Unless you know of some workaround that I'm not aware of

core/src/proof/normalization.rs
```diff
@@ -28,3 +32,10 @@ pub fn create_hashed_normalized_doc(
 
     Ok(output)
 }
+
+pub fn hash(data: &[u8]) -> Result<[u8; 32], crate::error::Error> {
```
gshuflin, line 36: Are any of these functions fallible? Why does this need to be `Result<_, Error>`?

LuisOsta: Good catch!

core/src/proof/normalization.rs
```diff
@@ -28,3 +32,10 @@ pub fn create_hashed_normalized_doc(
 
     Ok(output)
 }
+
+pub fn hash(data: &[u8]) -> [u8; 32] {
+    let mut hasher = sophia::c14n::hash::Sha256::initialize();
+    hasher.update(data);
+    let hash_data = hasher.finalize();
+    hash_data
```
gshuflin, line 40: We can also just return `hasher.finalize()` here

gshuflin (approved):

